PR #6, opened 2021-01-20
Camino-deployment-setup

taoteg:
- Updated the .gitignore to disregard the docker_repo.var file in the codebase. 
- Edited the Makefile to reflect the correct filename in the assignment block to docker_repo.var.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ build
 
 ## environment variables
 .env
+docker_repo.var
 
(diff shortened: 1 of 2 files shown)


taoteg:
This is really it for the CORE-cms side.
Now we need to setup Camino branches and Jenkins scripts.

tacc-wbomar (approved):

PR #189, opened 2021-04-12
task/FP-982: Add `rebuild_index` haystack signal processor

rstijerina:
### Overview
On a document publish, unpublish, or delete event, we want the cms index to rebuild. This adds a new signal processor that hooks into Haystack, and subclasses the [BaseSignalProcessor](https://django-haystack.readthedocs.io/en/master/signal_processors.html#default-basesignalprocessor) [class](https://github.com/django-haystack/django-haystack/blob/master/haystack/signals.py#L7).

### Notes:
An improvement would be to add/remove the document from the index instead of rebuilding, but I had trouble updating the index with the instance object returned from the `post_publish` signal. Since we have relatively few documents and activity on each cms, the occasional [rebuild_index](https://django-haystack.readthedocs.io/en/master/management_commands.html#rebuild-index) call will not be a resource burden.

### Testing:
1. Load the cms locally, and add an unpublished document.
2. Curl `localhost:9201/cms-dev-cms/_search` and verify the document is not in the index
3. Publish the document, curl `localhost:9201/cms-dev-cms/_search`, and verify the document is now there
4. Unpublish or delete the document, curl `localhost:9201/cms-dev-cms/_search`, and verify the document is gone

### DIFF
--- a/taccsite_cms/signal_processor.py
+++ b/taccsite_cms/signal_processor.py
@@ -1,0 +1,20 @@
+from haystack.signals import BaseSignalProcessor
+from django.db import models
+from django.core.management import call_command
+from cms import signals
+
+
+class RealtimeSignalProcessor(BaseSignalProcessor):
+
+    def setup(self):
+        signals.post_publish.connect(self.handle_save)
+        signals.post_unpublish.connect(self.handle_save)
+        models.signals.post_delete.connect(self.handle_save)
+
+    def teardown(self):
+        signals.post_publish.disconnect(self.handle_save)
+        signals.post_unpublish.disconnect(self.handle_save)
+        models.signals.post_delete.disconnect(self.handle_save)
+
+    def handle_save(self, **kwargs):
+        call_command('rebuild_index', '--noinput')
--- a/taccsite_cms/signal_processor.py
+++ b/taccsite_cms/signal_processor.py
@@ -5,6 +5,15 @@ from cms import signals
 
 
 class RealtimeSignalProcessor(BaseSignalProcessor):
+    """
+    A signal processor to make a call to the Django Haystack
+    `rebuild_index` management command when a CMS document is
+    published, unpublished, or deleted.
+
+    Usage:
+        This signal processor hooks into Haystack via the
+        `HAYSTACK_SIGNAL_PROCESSOR` Django setting.
+    """
 
     def setup(self):
         signals.post_publish.connect(self.handle_save)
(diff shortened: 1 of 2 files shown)


tacc-wbomar (approved):
Elegant.

__Testing__
- ✓ Confirm index did not auto-update on `main`.
- ✓ Confirm publish of edited page updates index.
    - "Text" plugin addition
    - "Text" plugin change
- ✓ Confirm page removal  updates index.
- ✓ Confirm page addition (after publish) updates index.

taccsite_cms/signal_processor.py
```diff
@@ -0,0 +1,20 @@
+from haystack.signals import BaseSignalProcessor
+from django.db import models
+from django.core.management import call_command
+from cms import signals
+
+
+class RealtimeSignalProcessor(BaseSignalProcessor):
```
tacc-wbomar, line 7: Please document for new users.

tacc-wbomar (changes requested):

taccsite_cms/signal_processor.py
tacc-wbomar: Maybe a `TODO: …` also for that improvement you mentioned in "Notes".

rstijerina: Done

taccsite_cms/signal_processor.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    A signal processor to make a call to the Django Haystack
+    `rebuild_index` management command when a CMS document is
+    published, unpublished, or deleted.
+
+    Usage:
+        This signal processor hooks into Haystack via the
+        `HAYSTACK_SIGNAL_PROCESSOR` Django setting.
+    """
```
tacc-wbomar, line 16: Would you please use the word "search" somewhere? Maybe

```suggestion
    """
    Calls the Django Haystack `rebuild_index` management
    command (to rebuild search index) when a CMS document
    is published, unpublished, or deleted.

    Usage:
        This signal processor hooks into Haystack via the
        `HAYSTACK_SIGNAL_PROCESSOR` Django setting.
    """
```

_It's a bit technical, which at least will keep people from breaking it. Perhaps they could bite down on `rebuild_index` and follow the trail. But just letting them now that this is search-related is the hope._

tacc-wbomar (commented):
I don't want to nitpick, but at least let Bob McNewguy know this is search-related. A haystack signal process who what, now? I get it, but I sure didn't used to.

tacc-wbomar (approved):
Thank you for mentioning search!

PR #203, opened 2021-04-22
Revert "Merge pull request #189 from TACC/task/FP-982"

tacc-wbomar:
# Goal

Remove code (that re-indexes search on page publish, un-publish, and delete) until we can find a solution to the real problem.

# Background

PR #189 was well-tested locally. But upon its deploy to ProTX—and possibly to Core and Neuronex, so suggests Sal's experience which matches John and Wes' on ProTX—it caused 500 errors on attempts to publish. Those 500 errors came from Elastic Search authentication issue. When, on ProTX server, John fixed some mistyped ES credentials in `secrets.py`, the problem __still__ existed. After hunting ES-related changes, PR #189 seemed most likely to trigger the issue. When that "trigger" code was removed, the problem (500s) went away.


tacc-wbomar:
**Declined.** The problem is ES authentication. This was a temp "fix", and it not helpful for Core nor Neuronex, which had a different problem and solution.

Issue #93, opened 2021-01-29
Article Lists Styles


tacc-wbomar:
# Goal

Style all types of [Article List](https://confluence.tacc.utexas.edu/x/OIAjCQ)s.

# Requirements

1. __Must__ have style for "thumbnail & type & date & title & abstract (as grid item)".*
1. __Must__ have style for "thumbnail & type & date & title & abstract (as list item)".*
1. __Must__ have style for "thumbnail & title & dates (as grid item)".
1. __Must__ have style for "date & title & summary".

> \*These may or may not get a byline. They were added to #95, and might be propagated (Matt is not sure, yet).

# Direction

- Styles for the "See All …" links come from #76.
- Styles for "Read more…" truncation come from #96.

# Reference

* [Frontera Redesign](https://xd.adobe.com/view/90e6b252-0293-4243-7748-11d94a0763d0-e6d7/grid)
    * home page news
    * science page news
    * news page news (grid)
    * news page news (list)
    * home page allocations
    * allocations page
    * home page events
    * events page events

PR #265, opened 2021-07-01
Quick: Clean Up Plugins to Match GH-93 Plugins

tacc-wbomar:
# Overview

Update old plugins to have formatting and naming consistent with incoming GH-93 plugins.

# Changes

- new lines
- organizational comments
- rename `utils.py` to `helpers.py`
- remove what would become redundant `documentation`
- use internationalization formatting for user-facing strings with variables
- improve & add docblocks


tacc-wbomar:
Self-merging because it's almost a no-op and I tested it and it's to dev.

PR #314, opened 2021-08-16
Hotfix: Restore variable deleted during GH-98

tacc-wbomar:
# Overview

Restore missing variable (deleted via https://github.com/TACC/Core-CMS/pull/312).

# Changes

- Restore variable.

# Screenshots

Before
Hotfix X-Dark RGB - Before

After
Hotfix X-Dark RGB - After

# Testing

Difficult to test because homepage is not quickly reproducible locally, but I have it.

# Notes

This variable is used by section object banner (from Frontera homepage redesign).

PR #312, opened 2021-08-13
GH-98: Typography

tacc-wbomar:
# Overview

Standardize font sizes and known colors, and known usage of them based on available designs.

# Changes

- __New__: Set site text color and font size.
- __New__: Set site heading styles.
- _Minor_: Do not assume `dt` font weight is bold.
- __New__: Set site link styles.
- _Minor_: Set text color and font size variables.
- _Minor_: Prevent new dropdown text bug.

# Issues

- #98

# Screenshots

GH-98 - Render Snippet (Style Guide)

# Testing

1. If no environment, then [prepare to test CMS changes](https://github.com/TACC/Core-CMS/wiki/Test-CMS-Changes).
2. Add new snippet to test page.
3. Create Snippet plugin with:
    - __HTML__: (none)
    - __Template__: `snippets/typography-guide.html`
    - [screenshot "GH-98 - Snippet Form"](https://user-images.githubusercontent.com/62723358/129333183-65ab88b6-dbf0-4b19-9b10-3f3a06f13fbe.png)
4. Test that the following is rendered with the template of your choice:
    - (Recommended) With page template "Style Guide"✱:
      [screenshot "GH-98 - Render Snippet (Style Guide)"](https://user-images.githubusercontent.com/62723358/129333186-1ffd72a2-e112-4a97-ada2-710d4f0c4821.jpeg)
    - With page template "Fullwidth":
      [screenshot "GH-98 - Render Snippet (Fullwidth)"](https://user-images.githubusercontent.com/62723358/129333501-abd11e67-adf1-4d4c-817e-48deea2604a9.jpeg)
5. Create sub-pages that show in menu so your CMS nav menu can have a dropdown.
6. Ensure CMS nav menu dropdown text is `16px`.
7. (Optional) Ensure Portal nav menu dropdown text is `16px`.†
8. (Optional) As you know how to do so, test that these pages are affected as described:
    - Frontera Homepage should have almost no change (header font may be thicker).‡
    - These pages have new body and heading text color:
    	- Standard Portal Homepage §
    	- any "Guide: …" template page §

> \* Choose template from "Page" > "Templates" menu. To add missing template, add relevant line from `default_secrets.py` to `secrets.py` (see `_CMS_TEMPLATES` entry).
[...]

Issue #98, opened 2021-01-29
Global Typography Styles & Fonts


tacc-wbomar:
# Goal

Standardize font sizes, families, files, styles, et cetera.

PR #316, opened 2021-08-16
Hotfix: Restore variable deleted during GH-98

tacc-wbomar:
# Overview

Restore (another) missing variable (deleted via https://github.com/TACC/Core-CMS/pull/312).

_I just did this in https://github.com/TACC/Core-CMS/pull/314, with a related variable._

# Changes

- Restore variable.

# Screenshots / Testing

**N/A**

_The variable is used in CSS, but no existing page yet renders its result._

# Notes

This variable would be shown with a light section object banner (from Frontera homepage redesign).


tacc-wbomar:
Merging because it's almost a noop. I am fixing a big visual bug caused by a small reason. The bug would only show up if user creates light section (which only I know how to do, until I make a section plugin).

Issue #101, opened 2021-01-29
Header Redesign → FP-1458


wesleyboar:
# Status

Migrated to [FP-1458](https://jira.tacc.utexas.edu/browse/FP-1458).

# Goal

- Update header style to match newest design.
- Use Cortal icons ([FP-526: Install Header Cortal Icons](https://jira.tacc.utexas.edu/browse/FP-526)).

# Direction

Unless requested, __avoid__ performing technical debt and nice to haves, like:
- https://github.com/TACC/Core-CMS/issues/12
- https://github.com/TACC/Core-CMS/issues/62
- https://github.com/TACC/Core-CMS/issues/55

The styles are not ideal, but there _are_ maintainable. The markup is duplicated, but the duplication is _much less_ than it was before you took over.

# Reference

- [(Confluence) Shared UI - Components - Header](https://confluence.tacc.utexas.edu/x/RIAjCQ) (see "Mid 2021")
- [(Adobe XD) CMS-Common Components](https://xd.adobe.com/view/b37b291a-7798-4814-90f3-7ac63c48a3e0-e628/grid)
- <details>
	<summary>If you need custom navbar transisitons instead of Bootstrap's…</summary>

    - [Turning off Twitter Bootstrap Navbar Transition animation](https://newbedev.com/turning-off-twitter-bootstrap-navbar-transition-animation)
    - [Responsive Animated Navbar with CSS](https://fireship.io/lessons/css-responsive-icon-navbar/)

	</details>

wesleyboar:
Migrated to JIRA as [FP-1458](https://jira.tacc.utexas.edu/browse/FP-1458) a long time ago.

PR #336, opened 2021-09-02
Hotfix: Remove errant comma in settings

tacc-wbomar:
# Goal

Remove errant comma in `settings.py`.

# Notes

I am surprised more branches aren't failing, given that this caused a server crash on `task/GH-101...`.


tacc-wbomar:
Emergency fix. Self-merging.

Issue #339, opened 2021-09-10
Support & Fix Cache Settings via Secrets


tacc-wbomar:
# Goal

1. Fix Django cache support via `secrets.py`.
2. Support CMS cache settings via `secrets.py`.

# Requirements

- Developer **should** be able to control cache via local settings.

# Direction

For __Goal 1__, see that `secrets` (undefined, outdated) is referenced instead of `current_secrets`:
https://github.com/TACC/Core-CMS/blob/1248117aa0521d331525f83d40d52c9a6126609d/taccsite_cms/settings.py#L318-L319
So, adding `_CACHE` to local secrets would crash the server.

# Background

Local settings (secret or not) are currently managed via local secrets. There is an issue to change this:
- https://github.com/TACC/Core-CMS/issues/39

# Reference

- [Django Caching](https://docs.djangoproject.com/en/2.2/ref/settings/#caches)
- [Django CMS Caching](https://docs.django-cms.org/en/release-3.7.x/how_to/caching.html#settings)

tacc-wbomar:
This can now be done with `settings_local.py`.

Issue #39, opened 2021-01-21
Separate Custom Settings From Secrets


tacc-wbomar:
# Goal

Extract non-secret custom settings from `secrets.py` so that they can be saved in the repo to reduce the manual maintenance of `secrets.py`.

_Core Portal will do the same for itself via [FP-304]. The "Direction" is based on that of [FP-304]._

# Requirements

1. Non-secret settings __must not__ be in `secrets.py`.
1. Non-secret settings __must__ be comited to the repo.

# Direction

## Portal's Solution

- https://github.com/TACC/Core-Portal/pull/393

## Original Idea

<details><summary>Archived</summary>

1. In the `{env}.env` file (e.g. `dev.env`) of the [TACC/Core-Portal-Deployments](https://github.com/TACC/Core-Portal-Deployments/) repo:

    ```
    CUSTOM_SETTINGS_CMS=dev.custom_settings_cms.py
    ```
1. In the deploy script for the [Jenkins "Core Portal Deploy" job](https://jenkins01.tacc.utexas.edu/job/Core_Portal_Deploy/):

    ```
    cp $CUSTOM_SETTINGS /opt/portal/Camino/conf/portal/custom_settings.py
    ```
3. In [TACC/Core-CMS](https://github.com/TACC/Core-CMS/) `settings.py`:

    ```
    from taccsite_cms.custom_settings import *
    ```
4. Use `custom_settings.py` instead of `.env` for `THEME`.
5. Remove `THEME` from [Jenkins "Core CMS" job](https://jenkins01.tacc.utexas.edu/job/Core_CMS/).
6. Migrate non-secret `settings.py` values to `custom_settings.py`.
7. Remove non-secret `settings.py` values from CMS Stache entries.
8. Update [TACC/Core-Portal](https://github.com/TACC/Core-Portal)'s `sample.secrets.py`.

</details>

# Background

In GitLab by @taoteg on Jul 22, 2020, 18:05

> Right now the branindg configurations for each portal is included in its secrets.py file, but these values are not sensitive (eg - secret).
> 
> In order to better clarify the role of the secrets.py file, and better clarify the location of branding specific configurations, separate all bradning and logo settings out into branding.py and then overwrite the default values (automatically configured using demo values as of the new settings update). This will result in the unique values (only) for each SAD CMS site living in the secrets.py file and overwriting default values living in settings.py, default_settings.py (formerly example_setting.py) and branding.py.

[...]

PR #340, opened 2021-09-10
GH-339: Support Local Django and Django CMS Cache Settings

tacc-wbomar:
# Important

- __Upcoming Compatibility Concern__: This PR may not be compatible with a major change coming to CMS architecture.
- __Known Better Settings Management Solution__: https://github.com/TACC/Core-CMS/issues/39

# Overview & Changes

Allow dev to disable (or otherwise manipulate) cache for (at least) local development.

1. Fix typo in settings that causes crash on trying to set Django cache via secrets.
2. Support Django CMS cache settings (in secrets, like Django cache is supported/fixed).

# Issues

- Closes GH-339

# Screenshots

N/A

# Testing

## 1. Fix Django Cache Setting Support

1. Add `_CACHE` to local `secrets.py` with [valid value](https://docs.djangoproject.com/en/2.2/ref/settings/#caches).
2. Ensure server does not crash when run.

## 2. Add Django CMS Cache Setting Support

1. Add `_CMS_PAGE_CACHE`, `_CMS_PLACEHOLDER_CACHE`, and `_CMS_PLUGIN_CACHE` to local `secrets.py` with valid values:
    - `True`
    - `False`
2. Ensure server does not crash when run.

### DIFF
--- a/taccsite_cms/settings.py
+++ b/taccsite_cms/settings.py
@@ -316,7 +316,7 @@ if LDAP_ENABLED:
     ''' End LDAP Auth Settings '''
 
 if getattr(current_secrets, '_CACHES', None):
-    CACHES = secrets._CACHES                        # Are we actually using this setting?
+    CACHES = current_secrets._CACHES
 
 DATABASES = current_secrets._DATABASES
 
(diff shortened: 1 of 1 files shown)


tacc-wbomar:
This is no longer necessary for Core settings, since #39. This code may be (and should be) added to `settings_local.py` for local and dev servers. Documentation has been added: https://github.com/TACC/Core-CMS/wiki/Prevent-Cache.

PR #341, opened 2021-09-13
Updating CMS to support login via Core Portal

xandroc:
To test this change CEPv2 will need to be pulled and built from the [branch](https://github.com/TACC/Core-Portal/tree/task/FP-1194). The instructions to run and test the combined CMS and CEPv2 changes are in the CEPv2 PR [here](https://github.com/TACC/Core-Portal/pull/494)


Makefile
```diff
@@ -1,4 +1,4 @@
-DOCKERHUB_REPO := taccwma/$(shell cat ./docker_repo.var)
```
taoteg, line 1: We need to keep this line using the configurable `./docker_repo.var` setting.

Why:
The value in that file is used when generating the dockerhub image name that is unique to each portal.

Details:
While all portals share a common `core-portal` container, the CMS container is unique per portal due to the embedded branding and custom navigation elements, templates, styles, etc. 

The string value in the`docker_repo.var` file is used in the CMS codebase to:
- correlate the contents of the branding submodule in the `Core-CMS-Resources` repo 
- determine the name of the published  dockerhub image and target repo

We generate the CMS dockerhub images using the nomenclature of `$PROJECT-cms`. If you look under the taccwma dockerhub repos (https://hub.docker.com/orgs/taccwma/repositories), you will see `a2cps-cms`, `protx-cms`, `neuronex-cms`, etc. which comes from the value placed in the local docker_repo.var` file during the image build phase of the Jenkins job (here under the `Build` section: http://jenkins01.tacc.utexas.edu:8080/job/Core_CMS/configure)

xandroc: Good catch! I'll update the Makefile to restore the functionality

taoteg (changes requested):
LGTM except for that one line fix in the Makefile. =)

taccsite_cms/urls.py
```diff
@@ -6,42 +6,37 @@
...
         url(r'^cms/nav/search/markup/$', TemplateView.as_view(template_name='nav_search.raw.html'), name='search_bar_markup'),
         url(r'^cms/nav/pages/markup/$', TemplateView.as_view(template_name='nav_cms.raw.html'), name='menu_pages_markup'),
         url(r'^cms/header/branding/markup/$', TemplateView.as_view(template_name='header_branding.html'), name='header_branding_markup'),
         url(r'^cms/header/logo/markup/$', TemplateView.as_view(template_name='header_logo.html'), name='header_logo_markup'),
-    ]
-
-if settings.FEATURES['blog']:
-    urlpatterns += [
-        # Support `taggit_autosuggest` (from `djangocms-blog`)
-        url(r'^taggit_autosuggest/', include('taggit_autosuggest.urls')),
```
tacc-wbomar, line 38: ⚠️ This removes a sub-feature of the CMS blog.

In the new setup, how should this addition to `urls.py` be supported?

_Related to removal of Blog from [`settings.py`](https://github.com/TACC/Core-CMS/pull/341/files#diff-16fd09f645fd22d5389d1711d21583bb9ce289bb09179cb8fa2b6fbf41aac385)._

taccsite_cms/settings.py
```diff
@@ -375,52 +357,6 @@ def get_subdirs_as_module_names(path):
...
-    )
-
-    # Blog/News: Change default values for the auto-setup of one `BlogConfig`
-    # SEE: https://github.com/nephila/djangocms-blog/issues/629
-    BLOG_AUTO_SETUP = True
-    BLOG_AUTO_HOME_TITLE ='Home'
-    BLOG_AUTO_BLOG_TITLE = 'News'
-    BLOG_AUTO_APP_TITLE = 'News'
-
-
```
tacc-wbomar, line 423: ⚠️ This change removes a large feature that is already installed on Texascale.

1. What is the plan for supporting this feature per site without duplicate setup?
2. What is the plan for preventing failure on Texascale when it loads this change?

<details><summary>Notes</summary>

- This feature is planned to be used for TACC, also.
- This is related to removal of a sub-feature in [`urls.py`](https://github.com/TACC/Core-CMS/pull/341/files#diff-ed9c8319f3a9f96a3b5db913871c815cf80ff229a6134b3655396a1e0a7e2701)

</details>

taoteg (approved):
Good catch on the auth fields! Looks solid.

taccsite_cms/settings.py
```diff
@@ -234,15 +287,10 @@ def getsecrets():
...
     'test_without_migrations',
     'taccsite_cms',
-    # Restore djangocms plugins that bootstrap4 hides
     'taccsite_cms.contrib.bootstrap4_djangocms_link',
     'taccsite_cms.contrib.bootstrap4_djangocms_picture',
-    # TODO: Extract TACC CMS UI components into pip-installable plugins
-    # FAQ: The djangocms_bootstrap4 library can serve as an example
     'taccsite_cms.contrib.taccsite_sample',
     'taccsite_cms.contrib.taccsite_system_monitor',
     'taccsite_cms.contrib.taccsite_data_list'
```
tacc-wbomar, line 296: ❓ Unexpected isolated removal of individual comments.

1. What is the logic for deciding which comments to retain?

<details><summary>Notes</summary>

- Other comments about intent or details of the code below (like the `get_subdirs_as_module_names` function comments just below in this file) have **not** been removed.
- I have found that unused settings (hints) seem to be consistently removed.

</details>

tacc-wbomar (commented):
I notice that `FEATURES` has been emptied and undocumented, and its `blog` feature removed. I have concerns and questions.

tacc-wbomar:
The changes on which I have commented ([review](https://github.com/TACC/Core-CMS/pull/341#pullrequestreview-754190986)) seem independent of FP-1194. How do they relate?

This PR → https://github.com/TACC/Core-Portal/pull/494 → [FP-1194: Synchronize User Login between Core Portal and CMS](https://jira.tacc.utexas.edu/browse/FP-1194)

taccsite_cms/settings.py
```diff
@@ -478,27 +394,37 @@ def get_subdirs_as_module_names(path):
...
+        'URL': ES_HOSTS,
+        'INDEX_NAME': 'cms',
+        'KWARGS': {'http_auth': ES_AUTH }
     }
 }
 
-ES_DOMAIN = current_secrets._ES_DOMAIN
+SETTINGS_EXPORT_VARIABLE_NAME = 'settings'
+DEBUG = True
+FEATURES = ''
```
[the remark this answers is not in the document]
xandroc: Yeah, I see what you mean, I left FEATURES here to prevent breaking spots where it's being referenced. I'll go back in future PR's and clean-up

taccsite_cms/default_secrets.py
```diff
@@ -1,305 +0,0 @@
-# TACC CMS SITE TEMPLATE SETTINGS.
```
rstijerina, line 1: do we need this to be deleted?

taccsite_cms/settings.py
```diff
@@ -1,104 +1,172 @@
...
+import ldap
+from django_auth_ldap.config import LDAPSearch, GroupOfNamesType
 
-
+SECRET_KEY = 'CHANGE_ME'
 def gettext(s): return s
 
+DATA_DIR = os.path.dirname(os.path.dirname(__file__))
 
-# Import secret values dynamically without breaking portal.
```
rstijerina, line 23: This seems like a major refactor. Could we pull out what is required by this PR from the rest of the refactor?

taccsite_cms/remote_cms_auth.py
```diff
@@ -0,0 +1,93 @@
...
+import requests
+import logging
+from django.conf import settings
+UserModel = get_user_model()
+
+logger = logging.getLogger(__name__)
+
+
+def verify_and_auth(request):
+    user = auth.authenticate(request)
```
rstijerina, line 15: Confirming - this will not throw an exception when it fails to authenticate?

taccsite_cms/remote_cms_auth.py
```diff
@@ -0,0 +1,93 @@
...
+    if user:
+        # User is valid.  Set request.user and persist user in the session by logging the user in.
+        request.user = user
```
rstijerina, line 18: is request.user not already user?

taccsite_cms/remote_cms_auth.py
```diff
@@ -0,0 +1,93 @@
...
+        try:
+            stored_backend = load_backend(request.session.get(auth.BACKEND_SESSION_KEY, ''))
+        except ImportError:
+            # backend failed to load
+            auth.logout(request)
+        else:
+            if isinstance(stored_backend, RemoteUserBackend):
+                auth.logout(request)
+
+    def configure_user(self, request, user):
```
rstijerina, line 87: Is this setup for future modification of the user?

rstijerina (changes requested):
Looking good. Can we isolate some of the changes into separate PRs?

taccsite_cms/urls.py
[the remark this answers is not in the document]
xandroc: Good question, thanks for asking how we can solve this. 

This actually affects the way the whole blog is set-up. Since this base repo is meant to support a number of CMS's (most of which won't need a blog) there are a couple of ways we can add custom functionality like this without using a lot of conditional logic in the base project. One option is to break out the blog into an app in it's own repo (I believe the orignal code for this is set-up that way so you can potentially fork that repo [or just reference that project in requirements and override what you need] and follow that same pattern) and then either publish the app to pypi and add a reference in requirements.txt or point to the repo directly in the requirements file. You can add the required urls.py changes in the taccsite_custom directory for any site that may (including INSTALL_APPS entries, etc) need the blog. 

Alternatively, if you don't want to create a new repo or think the blog code will only ever serve one site, you could move the code to the specific taccsite_custom dir and  make the changes there.

Personally, I think the first option is a great opportunity to fork a popular project and get something published to pypi that others may be able to use. 

Either way, the idea is to find solutions that centralize as little code as possible, and push out deviations from core so that features are configurable via deltas in each custom cms project instead of the core project having to be aware of every possible feature used in a child/custom implementation.

thoughts?

taccsite_cms/settings.py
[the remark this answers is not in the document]
taoteg: Can these settings be added back in on a per portal basis using the `settings_custom` or `settings_local`, `settings_secret` files?

taccsite_cms/urls.py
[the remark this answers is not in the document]
taoteg: I would be in favor of allowing this logic to remain in the CMS code under `urls` because this is a key idea that Wes designed features around to allow the CMS to dynamically grow as needed rather than having to over-engineer a solution up front that fits all sdizes (when those sizes are forever unknown and TBD). To remove this simple bit of logic now for the sake of unneeded (and unexpected) austerity is counter to the work he has been doing on the CMS.

Explanation:

When we split the CMS codebase off a few years back, we intended to let the CMS codebase evolve to handle all of the customization needs that CMD will encounter for site content across the various sites. We broke it apart from the Core-Portal so that it actually can inculcate all these kinds of edge case features. We can not know in advance what sites will want the blog feature (or future features built around the CMS) but we know at least one does and several probably will, given the design options that feature enables for CMD.

taoteg:
Fully second the request for multiple PRs now that I am reviewing the feedback from Sal and Wes. The `settings.py` changes in particular (anything beyond what is required to make the new login feature work correctly). There are a lot more implications there than I realized on my (far too cursory) review.

taccsite_cms/settings.py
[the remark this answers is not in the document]
xandroc: This can remain in use, we just need to find a solution that implements a modular architecture. We can always work with CMD to ameliorate any potential issues

taccsite_cms/urls.py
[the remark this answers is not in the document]
tacc-wbomar: <details><summary>Known CMS Usage of Blog: Texascale, TACC</summary>

- Texascale should (installed, unused but enabled for testing).
- TACC CMS will

</details>

<details><summary>Installable App for Blog: Cool, but likely not now.</summary>

- Cool. This isolates and commits support for the feature.
- But, this endeavor is too large, for me, given my current skill to do in a week.
- Side note: I don't think TACC customization will serve the community. But PRs to `djangocms-blog` could.

</details>

<details><summary>Within <code>taccsite_custom</code>: Feasible.</summary>

It would be "saving" code, not using it—just like the other settings stored in there. Unless, a `settings.custom` file in that repo's directories are intended to be applied with the upcoming architecture change.

</details>

<details><summary>Use <code>settings_custom.example.py</code>: I propose this.</summary>

Portal has this file. It supports saving example customization and documentation for them. If blog becomes worth having more than just these settings and a `urls.py` bit, we can explore an "Installable App for Blog".

</details>

taccsite_cms/remote_cms_auth.py
xandroc: Should not unless there is some non-standard exception. Django authors handle the most likely exceptions, (also the ones that provide the most flexibility), you can see the error handling here, https://github.com/django/django/blob/2.2.16/django/contrib/auth/__init__.py#L66

taccsite_cms/default_secrets.py
tacc-wbomar: I think not.

I propose it be cleaned up as multiple `...example.py` files, like Portal has. [A `settings_example.py` file specifically would resolve my blog concern (because I can document how to add it).](https://github.com/TACC/Core-CMS/pull/341#discussion_r708533776)

taccsite_cms/settings.py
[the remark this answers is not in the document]
tacc-wbomar: Well, now that you have provided and inspired [blog feature solutions](https://github.com/TACC/Core-CMS/pull/341#discussion_r708533776), its only [the documentation](https://github.com/TACC/Core-CMS/pull/341#discussion_r708533776) that is missing. That could go here or in a `settings_custom.example.py` (like Portal has).

taccsite_cms/settings.py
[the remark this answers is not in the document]
tacc-wbomar: I propose `settings_custom.example.py`, like Portal ([informal proposal](https://github.com/TACC/Core-CMS/pull/341#discussion_r708533776), see and open final bullet).

taccsite_cms/remote_cms_auth.py
xandroc: At this point, the user has been authenticated but I don't think the user has been added to the request (unless the user already had a session, in which case it gets destroyed). Even the base ModelBackend implementation doesn't seem to add it to the request, I don't think. https://github.com/django/django/blob/2.2.16/django/contrib/auth/backends.py#L16

taccsite_cms/remote_cms_auth.py
xandroc: Yes, I took it from another implementation in case we want to do things like add permissions, the way to do it should be readily evident here

tacc-wbomar:
I noticed inconsistent [changes in `taccsite_custom`](https://github.com/TACC/Core-CMS-Resources/compare/50882bb...5a6f0ec). Intentionally so? I dunno. I opened a PR:

- https://github.com/TACC/Core-CMS-Resources/pull/74

taccsite_cms/urls.py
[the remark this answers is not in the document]
xandroc: @taoteg 
Unfortunately that key idea runs counter to the entire django/requirements/pypi architecture of modular design, which is in place so projects don't become monoliths that require knowledge of all features. In order for the Core CMS to be able to grow dynamically, it needs to remain (as much as possible) feature agnostic--so it doesn't have references to features/plugins/apps unless they are universally required (like the auth functionality we're building in here). 

In the same way that Django need-not know about django-cms, nor django-cms about each plugin that's built for it, a modular architecture means we build plugins/apps that can be added to Core CMS via a config like requirements and settings.py (installed_apps, etc) and not via if/else conditionals.

This is the same pattern followed throughout all software design, and without it, software like Django would quickly become unmanageable, which is where we're headed with Core CMS if we don't start moving towards adopting the standard model and pattern that makes software flexible and reusable.. and that means the core cms code should not be aware of any "features" unless they are deemed as required for most if-not-all use cases, and even then, it may make sense to break out some of those if they become overly large or useful on their own.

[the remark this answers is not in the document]
tacc-wbomar: [Private discussion](https://tacc-team.slack.com/archives/G015BTGGLF5/p1631650173051100) has lead to agreement on the "Within taccsite_custom" solution—specifically `taccsite_custom:texascale-cms/`.

<details><summary>Reasons</summary>

- We need not add _all_ functionality to Core.
- Ideally, we have one implementation via a TACC blog plugin that overrides `djangocms-blog` only as necessary.
- Currently, only Texascale is using the blog. We can commit reference implementation in its `settings_custom.py`.

</details>

<details><summary>Drawbacks</summary>

- There is not one source of code (yet).
- Other domains do not know how to add blog.
    - I'll use documentation as tape.

</details>

taccsite_cms/settings.py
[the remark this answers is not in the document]
tacc-wbomar: Based on https://github.com/TACC/Core-CMS/pull/341#discussion_r708623599, I'll migrate all documentation and code to `taccsite_custom:texascale-cms/…`, implement blog for Texascale, and plan for a blog app.

taccsite_cms/settings.py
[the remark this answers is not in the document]
tacc-wbomar: - [Resolution to implement blog just for Texascale for now](https://github.com/TACC/Core-CMS/pull/341#discussion_r708623599)
- [Architectural reason to not have blog in Core](https://github.com/TACC/Core-CMS/pull/341#discussion_r708583857)

taccsite_cms/settings.py
tacc-wbomar: @xandroc I still believe some of these comments are useful. Where else do you suggest this information go?

```
    # For faster testing, disable migrations during database creation
    # SEE: https://stackoverflow.com/a/37150997
    'test_without_migrations',
```
```
    # Restore djangocms plugins that bootstrap4 hides
    'taccsite_cms.contrib.bootstrap4_djangocms_…',
```

taccsite_cms/default_secrets.py
tacc-wbomar: There is useful documentation in this file, like

- [CMS_TEMPLATES](https://github.com/TACC/Core-CMS/blob/1248117/taccsite_cms/default_secrets.py#L59-L94)
- [FEATURES](https://github.com/TACC/Core-CMS/blob/1248117/taccsite_cms/default_secrets.py#L123-L157)
- [BRANDING & LOGOS & FAVICON](https://github.com/TACC/Core-CMS/blob/1248117/taccsite_cms/default_secrets.py#L167-L237)

I am not saying it must be here, but documentation should not be deleted without replacement or explanation that we agree makes it moot.

tacc-wbomar (changes requested):
There are issues that should make this PR difficult to approve:
1. [documentation was removed](https://github.com/TACC/Core-CMS/pull/341/files#r708702700)
2. [incorrect type for default value of a setting](https://github.com/TACC/Core-CMS/pull/341#discussion_r708714521)
3. lack of test steps or results for settings refactor beyond new auth feature
    <details><summary>Suggestion</summary>

	Deploy sites that have relevant or unique settings; or an interesting state:
	- Frontera:
		- `INCLUDES_CORE_PORTAL=True`
		- `CMS_TEMPLATES`
	- Texascale:
		- `INCLUDES_CORE_PORTAL=False`
		- `FEATURES['blog'] = True`
		- `CMS_TEMPLATES`
		- manually deployed
	- A2CPS:
		- needs the “new auth feature”
	- ProTX:
		- running off a not-main branch

    </details>

I've also logged some unresolved concerns:
- [removal of informative comments](https://github.com/TACC/Core-CMS/pull/341/files#r708467557)

_I am willing to help with some documentation and some deploys._

taccsite_cms/settings.py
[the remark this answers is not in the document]
xandroc: There are no issues that make this PR difficult to approve. Features property will be removed. As we determined documentation will move to a settings_custom.example.py or to a Readme, the core settings file should remain as clean as possible.

taccsite_cms/settings.py
xandroc: This is not a major refactor (one file changed in the core project, same for custom sites), but it is a change that puts us in a position to easily remove optional features from the core cms and push those out to the custom sites that need them (Something we've already discussed as good and can be done with minimal change after this change).  

Again, this is very low risk change that will go through testing after review. Any changes in settings not accounted for here will be easy to identify and update in testing.

taccsite_cms/settings.py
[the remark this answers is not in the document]
tacc-wbomar: Ah, this issue is not an approval-blocking concern. Thanks 👍. I've moved it from top list to bottom list on my review.

Cool. I am looking forward to restored documentation. I was unaware or had forgotten you agreed to `settings_custom.example.py`.

.dockerignore
```diff
@@ -1,6 +1,9 @@
 # Version Control
 .git
 
+**settings_custom.py
```
rstijerina, line 4: Could you include a `settings_custom.example.py` file for reference? It would better help me understand what should exist in this file

xandroc: Yup, will do

tacc-wbomar:
I have attempted to test this, but need direction to succeed. Please, assist.

__Test Steps__

0. Login to `tup-cms` dev server.
1. Remove underscores from variables in `secrets.py`.
2. Deploy CMS to `tup-cms` dev server.
3. Notice [deploy failure](https://jenkins01.tacc.utexas.edu/job/Core_Portal_Deploy/505/console) because of hostname.
	- The hostname `core_cms_postgres` comes from [`taccsite_cms/settings.py`](https://github.com/TACC/Core-CMS/blob/78753bf/taccsite_cms/settings.py#L48).
	- The hostname has not changed in the `secrets.py` on the server from _its secret value_.
	- The server connected to the database successfully before this deploy.

</details>

__Questions__

1. How do the custom secrets get loaded by settings?
	- <details><summary>Prior Implementation</summary>

		_The code that __had__ loaded them was [`def getsecrets():`](https://github.com/TACC/Core-CMS/pull/341/files#diff-16fd09f645fd22d5389d1711d21583bb9ce289bb09179cb8fa2b6fbf41aac385L24)._

		</details>

2. Are [`taccsite_custom/…/settings_custom.py` files](https://github.com/TACC/Core-CMS-Resources/compare/50882bb...5a6f0ec) now configured to overwrite [`taccsite_cms/settings.py`](https://github.com/TACC/Core-CMS/blob/task/FP-1194/taccsite_cms/settings.py)?
	- <details><summary>Prior Implementation</summary>

		_The [`taccsite_custom/…/settings_custom.py` files](…) were __not__ configured to overwrite [`taccsite_cms/settings.py`](https://github.com/TACC/Core-CMS/blob/2498c27/taccsite_cms/settings.py)._

		</details>

.dockerignore
xandroc: done

taccsite_cms/default_secrets.py
tacc-wbomar: - This file has been recently added (374a279) as `taccsite_cms/settings_custom.example.py`.
- I have created a new issue to revisit documentation after this PR: #343.

taccsite_cms/settings.py
tacc-wbomar: Revisiting in #343.

taoteg (approved):
Got this working across several CMS sites (TUP, A2CPS, CEP).
Approving so we can move forward.

taoteg:
Going to just do this and test again while I am an insomniac...

tacc-wbomar:
Closes #39

PR #365, opened 2021-09-27
Hotfix: (UTRC-356) Avoid Grid Blowout

tacc-wbomar:
# Overview

Avoid "grid blowout" (like UTRC homepage text below banner not wrapping*).

> \* This could happen if its `.s-focus-list` items have text that is longer than available space.

# Changes

- Add missing import of custom media queries.\*
- Use `min-width` (via `minmax()` for to prevent grid cells from overflowing grid box.

> \* This did not cause a bug for Frontera, because the build process for its homepage stylesheet imported other stylesheets that imported custom media queries.

# Notes

- [CSS Tricks: Preventing a Grid Blowout](https://css-tricks.com/preventing-a-grid-blowout/)


tacc-wbomar:
Closing in favor of larger PR #368 (which references this PR in its changeset).

PR #311, opened 2021-08-13
GH-310: Breadcrumbs

tacc-wbomar:
# Overview

Add breadcrumbs to standard (a.k.a. Fullwidth template) pages.

# Issues

- GH-310

# Changes

- _Minor_: Update `trumps/README.md` to say some scopes enforce markup.
- __Major__: Add & Import styles.
- __Major__: Add & Load templates.

# Screenshots

| Description | Image |
| --- | --- |
| GH-310 (Full Width Template) | GH-310 - Fullwidth Template | text is large†<br />no breadcrumbs✱ |
| GH-310 (Standard Template)<br />(with GH-98) | GH-310 (with GH-98) | text is small† |

<details><summary>Outdated</summary>

| Description | Image |
| --- | --- |
| GH-310 (Standard Template)<br />(sans GH-98) | GH-310 | text is large† |

</details>

> ✱ Small text occurs with GH-98 code. Large text occurs without GH-98 code.
> † Because the fullwidth template is used by pages that do not expect breadcrumbs __and__ breadcrumbs are a new standard page feature __and__ all the fullwidth pages I have seen also want full height control (but I am not changing the template name, because that requires too much post-deploy work).

# Testing

1. If no environment, then [prepare to test CMS changes](https://github.com/TACC/Core-CMS/wiki/Test-CMS-Changes).
2. Create/Edit a page to use new Standard template.
    1. In `taccsite_cms/secrets.py`, to `_CMS_TEMPLATES`, add `('standard.html', 'Standard'),`.
    2. Choose "Standard" from Template in "Page" > "Templates" CMS admin menu.
3. Ensure breadcrumbs appear on page.
4. Ensure breadcrumbs and page content have padding to their left and right.
5. Ensure breadcrumb styles match the following criteria✱:
    1. The __links__, match those in [a relevant design](https://xd.adobe.com/view/90e6b252-0293-4243-7748-11d94a0763d0-e6d7/screen/ff0348ff-577d-450a-bef0-5326f3f7c705/specs/).
    2. The __character__ between entries is a __">"__ (not a "|").
    3. The __current page__ text is __not__ a link.
    4. The __current page__ text is bold text.
6. Set the page to not show in menu.
    - [How to Hide Page from Menu](https://user-images.githubusercontent.com/62723358/129304055-62885eae-8426-448b-b799-2c1817f91f82.png)
7. Ensure breadcrumbs __still__ appear on page.
8. Edit the page to use the Fullwidth template.
[...]

Issue #310, opened 2021-08-13
Breadrcumbs


tacc-wbomar:
# Goal

Implement breadcrumbs.

# Requirements

- Parent links **should** be generated by CMS.
- Breadcrumbs **should** be on all standard pages.\*
- Styles **must** match design on [Frontera Redesign](https://xd.adobe.com/view/90e6b252-0293-4243-7748-11d94a0763d0-e6d7/grid) pages (that have breadcrumbs).
- Styles **must** be re-usable.

\* Core Fullwidth template is standard. Core Portal Home template is not standard.

PR #370, opened 2021-09-28
Quick: Permit Root Contianer in Structure

tacc-wbomar:
## Overview

- Allow CMS editors to retain/add Container in page structure.

## Issues

Fixes problem caused by #310.

## Screenshots

…

## Testing

…

## Notes

0. <details>
	<summary>Background</summary>

	0. Most v2 CMS pages use Full Width template, which has no `.container`.
	1. #311 introduced Standard template that provides `.container` and breadcrumbs.
	2. @tacc-wbomar announced that Standard template should be used:
	    - This is how to get breadcrumbs, which are (usually) for all pages except home page.
	    - This meant that "Container" must not be in Structure (because it adds nested padding).
	    - This means page edits on all existing v2 sites as they being to use Standard.
	3. Today, Wes realized that this could be solved with CSS instead of page edits.

	</details>
1. <details>
	<summary>I am open to suggestion for better location, but <a href="https://confluence.tacc.utexas.edu/x/IAA9Cw" target="_blank">ITCSS organization</a> has rules.</summary>

	- The [`bootstrap.container.css`](https://github.com/TACC/Core-CMS/blob/e4ed6d7d/taccsite_cms/static/site_cms/css/src/_imports/components/bootstrap.container.css) is a Component stylesheet. A component does not permit Scope class, like `.s-breadcrumb`.
	- If `.s-breadcrumb` is not included to limit range of style cascade, then the structure can hide nested Containers (usually undesired) because the tell-tale sign of nested padding would be hidden.
	- The [`s-breadcrumb.css`](https://github.com/TACC/Core-CMS/blob/adef3a6/taccsite_cms/static/site_cms/css/src/_imports/trumps/s-breadcrumbs.css) is a Scope stylesheet. A scope only permits styling of the scope and its children.
	- Both `.s-breadcrumb` and `.container` are equally relevant, so the styles should not belong to one or the other. And ITCSS _intentionally_ does not provide stylesheets for this kind of mixing.

	So, I opted to put the styles exactly where the problem is.

	</details>

### DIFF
--- a/taccsite_cms/templates/base.html
+++ b/taccsite_cms/templates/base.html
@@ -66,10 +66,18 @@
     <div id="content-wrap" class="o-site__body">
       <!-- This block used regardless. -->
       {% block content %}
+
+      {# To allow extra Container in page structure by hiding extra padding #}
+      <style type="text/css">
+        .container > .s-breadcrumbs + .container { padding-inline: 0; }
+      </style>
+
+      {# To wrap breadcrumbs in Container without relying on CMS editor #}
       <div class="container">
         {% include "nav_cms_breadcrumbs.html" %}
         {% placeholder "content" %}
       </div>
+
       {% endblock content %}
       <!-- End of Block. -->
     </div>


taoteg (commented):
LGTM.

PR #402, opened 2021-11-10
Navbar Toggler Icon Color tweak based on Theme settings

taoteg:
Adds a new `s-header.css` selector that sets the navbar-toggle-icon color based on the selected `settings.THEME` value in the portal configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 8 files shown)


taoteg:
Going to enhance this PR a bit to make it more permanent and extensible across all selectors that might be impacted by THEME settings.

taoteg:
Okay, made some changes but I think its ready to merge now.
Unless we have a more specific long-term solution to implement?

tacc-wbomar (approved):

tacc-wbomar:
There is a long-term change to `icon-burger` and `color` instead of this, but this is so quick and reliable that I don't see the need to wait. The `icon-burger` and `color` can come with header redesign (which I will break out into pieces for next year).

PR #418, opened 2021-12-08
FP-1377: New `u-text` Utility Class

tacc-wbomar:
### Closed

Opted for `<big>` via WYSIWYG instead of a utility class. If the need arises for a utility class, we can revive this PR.

<details>

I tired the tedious use of `<big>`. The font size is `19.2px`, which is close to `20px`. More importantly, using `<big>` is more intuitive (via WYSIWIG) for non-technical CMS editors than would be adding a class in "Advanced settings".

</details>

### Overview

Support larger text (except headings) with a utility class.

### Issues

[FP-1377](https://jira.tacc.utexas.edu/browse/FP-1377)

### Changes

* Added a new utility file with one class.
* Load the a file so class is available globally.

### Testing

1. Load [Frontera 2022 About Page (dev)](https://dev.fronteraweb.tacc.utexas.edu/2022/about/).
2. Inspect markup of the large text on the left side of page.
3. Confirm that the paragraph font size **is** `16px` because of `u-text--large`[^1].
4. Confirm that the headings' font size is set by code that overwrites the cascade of `u-text--large`[^1].

[^1]: The element with that class is `<div class="col col-lg-6 u-text--large">`.

### Notes

Bootstrap does not have a [utility class](https://getbootstrap.com/docs/4.0/utilities/text/) to do this, probably because a user could use `<big>`.[^2]

[^2]: I did not propose `<big>`, because there is not a maintainable way to ensure the text size is one of the [standard font sizes](https://confluence.tacc.utexas.edu/x/nh4FDg).


tacc-wbomar:
Closed in favor of WYSIWIG and `<big>`. Re-evaluation is welcome.

PR #437, opened 2022-01-27
Task/fp 1285 cms critical dependency upgrade

taoteg:
## Overview: ##

Bumping the base python image version and Pillow to eliminate a critical security vulnerability.

## Related Jira tickets: ##

* [FP-1285](https://jira.tacc.utexas.edu/browse/FP-1285)

## Testing Steps: ##

1.  Build and run CMS with no errors.

### DIFF
--- a/taccsite_cms/settings.py
+++ b/taccsite_cms/settings.py
@@ -17,6 +17,7 @@ from django_auth_ldap.config import LDAPSearch, GroupOfNamesType
 SECRET_KEY = 'CHANGE_ME'
 def gettext(s): return s
 
+
 DATA_DIR = os.path.dirname(os.path.dirname(__file__))
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
@@ -31,7 +32,7 @@ ALLOWED_HOSTS = ['0.0.0.0', '127.0.0.1', 'localhost', '*']   # In development.
 LDAP_ENABLED = True
 
 # Default portal authorization verification endpoint.
-CEP_AUTH_VERIFICATION_ENDPOINT =  'localhost' #'https://0.0.0.0:8000'
+CEP_AUTH_VERIFICATION_ENDPOINT = 'localhost'  # 'https://0.0.0.0:8000'
 
 ########################
 # DATABASE SETTINGS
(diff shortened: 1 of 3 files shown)


taccsite_cms/settings.py
```diff
@@ -167,14 +168,14 @@ def gettext(s): return s
     "img_file_src": "site_cms/img/favicons/favicon.ico"
 }
 
-INCLUDES_CORE_PORTAL = True
+INCLUDES_CORE_PORTAL = False
```
tacc-wbomar, line 171: **Uh Oh**: If this is set to `False` here, then all submodules that assume it is true now need to set it to `True`.

tacc-wbomar (commented):
Based on your research, I suggest docker base image be changed from to Python 3.7 to reduce potential friction. (I found `3.7(.12)-buster`) at https://hub.docker.com/_/python/.

<details>
<summary>Details</summary>

As you showed me in a [Pillow/Python dependency map](https://user-images.githubusercontent.com/62723358/151477666-b77b1ac8-1a86-4a38-a7c3-da32e33570ab.png), Pillow to v9 seems unsupported by our Python 3.6. So, the docker base image should change to Python 3.7 or 3.8. I don't foresee an issue, because we use Django 2.2.16 [which supports Python 3.5–3.8 (but not 3.9)](https://docs.djangoproject.com/en/3.2/releases/2.2/#python-compatibility).

</details>

taccsite_cms/settings.py
taoteg: Oops!! My bad, dev cruft. Thanks for the catch! Will fix and update this.

taoteg:
Thank you for the python <--> django catch!
That was what I was unsure of.
Eventually (*after* the Poetry conversion), I would like to bump `all teh things` to the latest versions we can support in the CMS container.
I managed to avoid that task for now, hopefully Poetry makes it far more trivial.
Updating the PR to fix the python version.

rstijerina (approved):
LGTM

tacc-wbomar:
_P.S. I had dependency error locally, but CMS—with this change—builds and deploys just fine with Camino ([build](https://jenkins01.tacc.utexas.edu/view/WMA%20CEP/job/Core_CMS_Build/359/),  [deploy](https://jenkins01.tacc.utexas.edu/job/Core_Portal_Deploy/1041))._

PR #443, opened 2022-02-16
Quick: Texascale: robots.txt 2022 / do not index 2022 pages

tacc-wbomar:
## Overview

On Texascale, do not index 2022 pages.


tacc-wbomar:
Simple change I do every year (but is much easier since last year). Self-approving and merging.

PR #465, opened 2022-03-12
FP-1544: Annotate Lightweight Tags

wesleyboar:
## Overview

Add script to annotate lightweight tags.

## Related

* [FP-1544](https://jira.tacc.utexas.edu/browse/FP-1544)

## Changes

- Add script.
- Ignore its output (used to manually check if script worked).

## Testing

1. Run the new script.
2. Confirm no errors.
3. Compare the two output files.
4. Confirm the "after" output difference has:
    - two dates per entry
    - no other content is changed

The whitespace in output files is different… sorry 🤷‍♂️ .  Ideas?

## Notes

After this process is approved, only **then** should we actually push the tags to origin. (Otherwise, origin might lose data from incorrect solution.)


bin/git_annotate_existing_tags.sh
```diff
@@ -0,0 +1,56 @@
...
+# Bug: (2022-03-12) (for a commit after v3.15.0)
+#      Running `git describe` gave unexpected output.
+#      - expect: "v3.15.0-…-g…"
+#      - actual: "v3.7.0-…-g…"
+# Why: 1. Most tags after (and before) v3.7.0 were not annotated. Similar:
+#         https://github.com/Reference-LAPACK/lapack/issues/123
+#      2. Because GitHub uses lightweight tags. Feedback for GitHub:
+#         https://github.com/github/feedback/discussions/4924
+# Fix: Run this script to retroactively annotate lightweight tags.
+#      (Note, this script also re-annotates annotated tags.)
```
tacc-wbomar, line 12: Add link to alternative solutions.

```suggestion
#      (Note, this script also re-annotates annotated tags.)
# Alt: See answers to https://stackoverflow.com/q/21738647/11817077
#      (These solutions are more elegant but might not preserve all lines, including new lines, in tag message.)
```

bin/git_annotate_existing_tags.sh
```diff
@@ -0,0 +1,58 @@
...
+# Alt: See answers to https://stackoverflow.com/q/21738647/11817077
+#      (These solutions are more elegant but might not preserve all lines, including new lines, in tag message.)
```
tacc-wbomar, line 14: Keep it under 80 chars.
```suggestion
#      (More elegant but may not preserve all of tag message with new lines.)
```

rstijerina:
How do you propose we make sure tags are added moving forwards in the release process?

wesleyboar:
Manually run this script upon each release. Not great, I know.

I really feel this is a problem GitHub should solve ([discussion](https://github.com/github/feedback/discussions/4924), [issue](https://github.com/isaacs/github/issues/1916)).

wesleyboar:
Proposal declined.

It's not worth changing our build process to accommodate this:
- benefit is a string in a comment on for devs who will rarely see it
- it's a problem GitHub should solve
- we want to move to a new build process anyway

wesleyboar:
This is apparently already part of `main` since v3.12.0-beta.4: e043d450.

bin/git_annotate_existing_tags.sh
wesleyboar, line 1: This script was eventually restored. It's available in `main` currently. [Permalink for the v4.9.0 one.](https://github.com/TACC/Core-CMS/blob/v4.9.0/bin/annotate-tag.sh)

PR #466, opened 2022-03-15
ECEP-113: News/Blog (Core, ECEP, Texascale)

wesleyboar:
## Overview

- In Core: install, configure, and style blog/news.
- Per project: configure blog/news.

## Related

- [ECEP-113](https://jira.tacc.utexas.edu/browse/ECEP-113)
- requires & required by https://github.com/TACC/Core-CMS-Resources/pull/127
- requires https://github.com/TACC/Core-Styles/pull/16

## Changes

- Improve relevant user-facing text.
- Add new settings (TACC Blog, Blog, Picture, Locale).
- Remove `FEATURES` settings.
- Install [meta](https://django-meta.readthedocs.io/en/latest/) and [django-page-meta](https://djangocms-page-meta.readthedocs.io/en/latest/usage.html) by default.﹡
- Style the blog.
- Support Picture alignment.
- Add CSS selector for CMS login state.
- Clean up stylesheet loading.
- Clean up Blog & Social Media markup loading and readmes.
- Customize Blog markup.
- Add snippet to hide main news image.

<sub>﹡ These had been bundled with Blog/News. But metadata should always be present; Blog/News is optional.</sub>

## Screenshots

Please see "Testing" → "Remote" instead.

### Remote

| Site | Dev. / Staging | Production |
| - | - | - |
| ECEP | [installed](https://pprd.ecep.tacc.utexas.edu/admin/djangocms_blog/blogconfig/) | [installed, configured, populated](https://prod.ecep.tacc.utexas.edu/news/) |
| Texascale | old install (unused) unable to update yet[^1] | old install (unused) unable to update yet[^1] |

[^1]: Texascale update pending [CEPWMA-679](https://jira.tacc.utexas.edu/browse/CEPWMA-679). Texascale might _still_ **not** use blog, because new TACC will take all of my bandwidth.

1. Verify `djangocms_blog` app is installed and active.
    - Open CMS Pages admin UI. New page has 🧩 icon?﹡
    - Open Blog/News page while logged in. Admin toolbar has "Blog" link?
    - Open CMS Administration UI. Section "DJANGO CMS BLOG" exists?

    ﹡ In same column where Home page has 🏠 icon.
2. Verify new Blog instance is working.
    - Add and publish an article. Article shows up on new page?﹡
    - Open new page while logged in. Page shows "No articles found"?

    ﹡ If article is not published, it will not show.
3. Verify that news article [listing](https://prod.ecep.tacc.utexas.edu/news) & pages (click articles in listing)…
[...]

PR #278, opened 2021-07-12
Quick: Support Non-Bootstrap Link & Picture Plugins

tacc-wbomar:
# Overview

Restore support for Generic plugins that `djangocms_bootstrap4` removes.

<details><summary>Why?</summary>

Most of the time, the extra features of Bootstrap4's link and image are:
1. unused by users
2. undesired by designers
3. add unnecessary complexity to development of TACC plugins that need simple image and link feature

</details>

# Changes

- Register new TACC plugin to re-register `LinkPlugin` (if Bootstrap's respective plugin is present).
- Register new TACC plugin to re-register `PicturePluign` (if Bootstrap's respective version is present).

# Screenshots

__Note__: I propose we mark the relevant Bootstrap plugins as deprecated.

<details>

Though Bootstrap plugins are still available (to support existing usage), they will __not__ be allowed to be nested within TACC plugins. Also, TACC plugins will extend Generic plugins, not Bootstrap plugins.

</details>
<details><summary>Legend</summary>

- Blue plugins are for links.
- Yellow plugins are for images.
- Outlined Bootstrap plugins are undesirable.
- Outlined Generic plugins are undesirable.

</details>

| Description | Image 1 | Image 2 |
| --- | --- | --- |
| | Link | Image |
| Add Plugins | Link Add | Image Add |
| Generic Plugin Form | Generic Link Form | Generic Image Form |
| Bootstrap Plugin Form | Bootstrap Link Form | Bootstrap Image Form |
| | Not Within "Text" Plugin | Within "Text" Plugin |
| Render Plugins | Bootstrap   Generic - Render - Not Within Text | Bootstrap   Generic - Render - Within Text |
| Plugin Structure | Bootstrap   Generic - Structure - Not Within Text | Bootstrap   Generic - Structure - Within Text |

# Testing

You may test locally or on servers that have deployed this code.

> __Last Recorded Deploys__
> - [Frontera (dev)](https://dev.fronteraweb.tacc.utexas.edu/) was [Build #454 (Aug 9, 2021 1:02:37 PM)](https://jenkins01.tacc.utexas.edu/job/Core_Portal_Deploy/454/) via `site/main/frontera` (which has this branch and more)
> - [Core (dev)](https://dev.cep.tacc.utexas.edu/) was [Build #453 (Aug 9, 2021 1:04:25 PM)](https://jenkins01.tacc.utexas.edu/job/Core_Portal_Deploy/453/) via `site/main/frontera` (which has this branch and more)

[...]

PR #474, opened 2022-03-30
Quick: Un-Deprecate Bootstrap Picture

wesleyboar:
## Overview

Do **not** warn users about using Picture plugin (and clarify when to use).

_The default responsive behavior of Bootstrap's image is desirable for some content._

<details><summary>Background</summary>

1. During #278, the warning was added.
2. Since then, I have accepted Bootstrap as part of the CMS.
3. During [ECEP-113](https://jira.tacc.utexas.edu/browse/ECEP-113), I realized the use case for Bootstrap picture.

</details>

## Related

* expected by #466

## Changes

- Remove deprecation icon and text.
- Change name of plugin to reflect intended usage.

## Testing

You can test on [dev.cep](https://dev.cep.tacc.utexas.edu/test/quick-un-deprecate-bootstrap-picture/) or locally.

1. Create page.
4. Add image via Bootstrap Picture plugin.
    - Verify **no** ⚠️ icon **nor** deprecation text.
    - Verify new name "Picture / Image (Responsive)"
5. Add image via Django Image plugin.
6. Both images should load fine.

## Screenshots

<details open><summary>What is the Change for Users?</summary>

| | Before | After |
| - | - | - |
| Search for Plugin | Before - Plugin Search | After - Plugin Search |
| Add the Plugin | Before - Plugin Create | After - Plugin Create |

</details>

<details><summary>Why Support Bootstrap Picture?</summary>

1. Bootstrap Picture applies the class `img-fluid` by default, and has a clear setting—a checkbox labeled "Responsive".
    Users don't know they usually want this. For devs, it's easier to allow Bootstrap Picture than customize the generic Picture plugin.

    | With "Responsive" Setting / `img-fluid` Class | Sans "Responsive" Setting / `img-fluid` Class |
    | - | - |
    | With img-fluid Class | Sans img-fluid Class |
    | ✅  image fits in text/content box | ❌  image escapes text/content box |

3. Django "Image" (above) is not responsive, so it can exceed container width (left). Bootstrap "Picture" (below) is responsive by default, so it will never exceed container width (left). If container is wider than images, neither image exceeds container width (right).

    | 1200px | 991px |
    | - | - |
    | 991px | 1200px |
    | window width < image width<br />Bootstrap "Picture" is only as wide as container | window width > images width, same size<br />image is only as big as intrinsic size |

</summary>


taoteg (approved):
LGTM.

PR #475, opened 2022-04-06
Quick: Remove Unused JavaScript

wesleyboar:
## Overview

Remove cruft JavaScript files.

## Related

none

## Changes

- Remove `site.js` and call to load it.
- Remove `nav_portal.js` (long since deprecated and never loaded).

## Testing

1. Confirm that this console log does not appear:
    ```js
    Core \`site.js\` has no logic.
     ```
2. Confirm that `site.js` is not loaded.

## Screenshots

No Console Log

No Site js Load


taccsite_cms/static/site_cms/js/nav_portal.js
```diff
@@ -1,51 +0,0 @@
...
-      document.getElementById('auth-menu').classList.remove('d-none');
-    },
-    function outputRequest(xhr, isFatal) {
-      if (isFatal) console.error(xhr);
-
-      // FAQ: Show the un-auth'd menu
-      document.getElementById('unauth-menu').classList.remove('d-none');
-    }
-  );
-})();
```
wesleyboar, line 51: **Explanation:** This whole file is unused. Instead of manipulating duplicated HTML, we now load the server-side–prepared HTML directly into place (https://github.com/TACC/Core-CMS/blob/main/taccsite_cms/templates/nav_portal.html#L16).

taccsite_cms/static/site_cms/js/site.js
```diff
@@ -1,3 +0,0 @@
-'use strict';
-
-console.info('Core `site.js` has no logic.');
```
wesleyboar, line 3: **Original Intent:** Note to self that there is a `site.js` for site-wide JavaScript. _But no site-wide JavaScript has since been needed._

wesleyboar (commented):
Notes for reviewers.

rstijerina (approved):

PR #485, opened 2022-05-05
Improve handling of missing custom settings

nathanfranklin:
## Overview

Altering the bare except when importing custom files to ModuleNotFoundError so that any issues with existing custom files can be seen.

## Testing

1. Put an error in one of the optional custom files (i.e. settings_custom.py secret.py or settings_local.py)
2. Ensure that error is raised

### DIFF
--- a/taccsite_cms/settings.py
+++ b/taccsite_cms/settings.py
@@ -507,21 +507,18 @@ META_USE_SCHEMAORG_PROPERTIES = True
 
 try:
     from taccsite_cms.settings_custom import *
-except:
-    None
-    # do nothing
+except ModuleNotFoundError:
+    pass
 
 try:
     from taccsite_cms.secrets import *
-except:
-    None
-    # do nothing
+except ModuleNotFoundError:
+    pass
 
 try:
     from taccsite_cms.settings_local import *
-except:
-    None
-    # do nothing
+except ModuleNotFoundError:
+    pass
 
 SETTINGS_EXPORT = [
     'DEBUG',


wesleyboar:
Thanks. Sorry for not understanding this simple improvement earlier. My Python comes and goes, net gain is gradual.

wesleyboar (commented):
I tested editing my `settings_local.py`.

1. I put a syntax error in. An error was raised. Got `SyntaxError: invalid syntax`. ✅
2. I used a non-existent variable. Got `NameError: name 'NO_VAR' is not defined`. ✅
3. I renamed `settings_local.py` file to `settings_local_badname.py`. **No** error. ✅
4. On `main` I repeated all the tests. **No** errors (meaning this PR fixes a problem). ✅ 

But I have questions:

1. What does `except: ModuleNotFoundError` do?
2. How do I actually get a `ModuleNotFoundError`?
    _Note: I would **not** expect one if a `settings_*.py` is missing, cuz those are optional._
3. Am I correct to see a disconnect between `ModuleNotFoundError` and the fact that I can get other errors?

nathanfranklin:
https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError

> Note: I would not expect one if a settings_*.py is missing, cuz those are optional.

If the files are missing, then a ModuleNotFoundError is thrown when we try to import.

There was a bare except before when trying to import those optional files so that we could continue if importing failed as they weren't t. ere. But that except is too broad and catches anything that goes wrong.  So now we just catch ModuleNotFoundError which is what we expect if the files aren't there. 

We don't catch any other exceptions as we want there is something wrong with our custom files and we want to know about those.

wesleyboar:
I got private guidance. I did not register `pass` as I read the code. I understand now that the `ModuleNotFoundError ` error is caught, but we `pass` on it, and other errors are not `pass`-ed, which is why I see them.

wesleyboar (approved):

wesleyboar:
This PR introduced code that I eventually learned from to perform #852. Thanks again, @nathanfranklin for it.

nathanfranklin:
> This PR introduced code that I eventually learned from to perform #852. Thanks again, @nathanfranklin for it.

awesome! @wesleyboar

PR #489, opened 2022-05-15
FP-1648: Update for Core-Styles API Change

wesleyboar:
## To Do

- [x] Pin `@tacc/core-styles` to commit or version, not [branch](https://github.com/TACC/Core-CMS/pull/489/files#r881827815). 66c6d14

## Overview

Update build command since [TACC/Core-Styles](https://github.com/TACC/Core-Styles) API change.

## Related

- [FP-1648](https://jira.tacc.utexas.edu/browse/FP-1648)
- requires https://github.com/TACC/Core-Styles/pull/23

## Changes

- __change `--input-dir` and `--output-dir` to `--input` and `--output`__﹡
- load branch of https://github.com/TACC/Core-Styles/pull/23
- _remove extra whitespace_

<sub>﹡ This makes it clearer that a user can specify individual files.</sub>

## Testing

```
npm ci
npm run build
```

1. See how many stylesheets are built on `main` branch.
2. See how many stylesheets are built on __this__ branch.
3. Verify same number of files is output as before this PR.

## Screenshots

N/A


package.json
```diff
@@ -9,7 +9,7 @@
     "npm": "8.5.5"
   },
   "devDependencies": {
-    "@tacc/core-styles": "git+https://git@github.com/TACC/Core-Styles.git",
+    "@tacc/core-styles": "git+https://git@github.com/TACC/Core-Styles.git#task/fp-1648-build-all-files-and-commit-dist",
```
rstijerina, line 12: I assume this branch hash will be removed before merge

rstijerina (approved):
LGTM

package.json
wesleyboar: Yes. Thanks. I need to think of a better way to do this for PRs (this is helpful for testing). I could use commits instead, like https://github.com/TACC/Core-CMS/pull/493#discussion_r880921494. Thoughts?

taoteg (approved):
LGTM. 
Making the same assumption about the branch hash removal before merge.

package.json
wesleyboar: Pinned to tag `v0.5.0` (66c6d14).

Long-term, I guess I need to:
- **either** version often (easier if I can independently test core-styles)
- **or** point to commits

PR #499, opened 2022-06-09
feat(tup): no LDAP

wesleyboar:
## Overview

Unset LDAP setting for TUP CMS.

## Related / Testing / Screenshots

- https://github.com/TACC/Core-CMS-Resources/pull/143

## Changes

- use Core-CMS-Resources commit that updates setting for tup-cms


taoteg:
Just noting that the when LDAP is set to true, the django container was able to force user login against the CMS container so the CMS could enforce page access permissions against individual users. Does the SSO solution take into account the CMS requirements for this?

wesleyboar:
@taoteg, I am not sure. @jchuahtacc, please see [his question](https://github.com/TACC/Core-CMS/pull/499#issuecomment-1151665555).

PR #504, opened 2022-06-27
fix: load fa icons even if no portal

wesleyboar:
## Overview

Support icons even if Portal is not active.

## Related

- [FP-1706](https://jira.tacc.utexas.edu/browse/FP-1706)

## Changes

- in template, remove portal conditional around icon load

## Testing

0. Have setting `INCLUDES_CORE_PORTAL` set to `False`.
1. Use a Font Awesome icon on a page.
2. Verify icon shows.

## Screenshots

FP-1706

## Notes

We do not want FA icons, but there is at least one usage.

Usage: Frontera homepage, "(→) See All News"


wesleyboar:
I keep loosing tiny bits of time to this bug, because I forget why it happens, poke around, realize why it happens, then never fix it.

PR #517, opened 2022-07-20
Support custom apps through config.

jarosenb:
## Overview
Add support for custom apps through `custom_app_settings.py` and `urls_custom.py`

## Related

- [TUP-250](https://jira.tacc.utexas.edu/browse/TUP-250)

## Changes

- Add configurable settings to insert new apps, middleware and static file paths into the core CMS.

## Testing

1. Copy `custom_app_settings.example.py` into `custom_app_settings.py` adn `urls_custom.example.py` into `urls_custom.py`
2. Start the CMS according to the README instructions and navigate to http://localhost:8000/test/. You should see the `custom_example` app being rendered.

## Screenshots

image


jchuahtacc (commented):
Does it make more sense to move `apps/custom_example` to a separate sample repository that has a Dockerfile that demonstrates a downstream repository's structure? We could approve this PR and build a hashed image before building the sample repository. So the Dockerfile in the other repo might be:

```
FROM tacc/core-cms:tag_for_this_PR
COPY /apps/custom_example /apps
```

etc.

taccsite_cms/custom_app_settings.example.py
```diff
@@ -0,0 +1,3 @@
+CUSTOM_APPS = ['apps.custom_example']
+CUSTOM_MIDDLEWARE = []
+STATICFILES_DIRS = ()
```
jchuahtacc, line 3: Is there a way for this file to also import settings values into the django settings object?

jarosenb: Not sure I understand the question- are you saying you'd prefer to do `from taccsite_cms.custom_app_settings import *` in the main settings file and mutate the arrays/tuples inside this file instead? Otherwise, there are already ways to override settings using `secrets` or `settings_custom`.

wesleyboar (commented):
I see, with these changes, how Core-CMS-Resources need **not** be a submodule.

<details><summary>Steps</summary>

1. Each Core-CMS-Resources directory becomes a repo or moves into `tup-ui:/libs`.
2. A project [`urls.py`](https://github.com/TACC/Core-CMS-Resources/blob/main/ecep-cms/urls.py) and [`settings_custom.py`](https://github.com/TACC/Core-CMS-Resources/blob/main/ecep-cms/settings_custom.py) would be loaded as a "custom app".
3. To load project static files, a CMS [`STATICFILES_DIR`](https://docs.djangoproject.com/en/2.2/ref/settings/#staticfiles-dirs) entry would point **outside** CMS.

    <details><summary>Code Change</summary>

    change
    https://github.com/TACC/Core-CMS/blob/f65d0c4e803b98b138ad76d0e014ae479088a2a5/taccsite_cms/settings.py#L215-L217

    to
    ```
        os.path.join(PATH_TO_THE_PROJECT_TO_RUN, 'static')
    )
    ```

    </details>

4. Which then means to locally run a project, Python needs access to project name.

    <details><summary>Ideas</summary>

    - Developer can add a local `docker_repo.var`, the same file [Jenkins config](https://jenkins01.tacc.utexas.edu/job/Core_CMS_Build/configure) uses.
    - We can use a `.env`, and update [Jenkins config](https://jenkins01.tacc.utexas.edu/job/Core_CMS_Build/configure) to read it instead.

    </details>

5. To load project [`templates/`](https://github.com/TACC/Core-CMS-Resources/tree/main/frontera-cms/templates), a CMS [`TEMPLATES`](https://docs.djangoproject.com/en/2.2/ref/settings/#staticfiles-dirs)`DIRS` entry would point **outside** CMS.

    <details><summary>Code Change</summary>

    change
    https://github.com/TACC/Core-CMS/blob/f65d0c4e803b98b138ad76d0e014ae479088a2a5/taccsite_cms/settings.py#L228-L230

    to
    ```
            'DIRS': [
                os.path.join(PATH_TO_THE_PROJECT_TO_RUN, 'templates')
    ```

    </details>

    _Not as proper as [a project cms app (poc)](https://github.com/wesleyboar/django-cms-quickstart--custom/tree/task/FP-1487-from-submod-to-apps) building off a [a core cms app (poc)](https://github.com/wesleyboar/django-cms-quickstart/tree/task/FP-1487-from-submod-to-apps)._

</details>

taccsite_cms/custom_app_settings.example.py
jchuahtacc: Sorry, I forgot there was a settings_custom.

jchuahtacc:
> 

We do not necessarily need them to be the same repo. They could, in theory, exist in the same Core-Cms-Resources repo where each directory has its own Dockerfile. It would require a little bit of tweaking to the Jenkins build job but would otherwise look the same from a Jenkins user's perspective.

wesleyboar:
> They could, in theory, exist in the same Core-Cms-Resources repo where each directory has its own Dockerfile. It would require a little bit of tweaking to the Jenkins build job [...]

Then I am excited. That is something I would like to see and learn from. I thought our Jenkins could not clone multiple repos in one job, thus could not mix and match content to build from, so I gave up on a Jenkins solution.

jchuahtacc (approved):

apps/custom_example/views.py
```diff
@@ -0,0 +1,9 @@
+from curses.ascii import HT
```
wesleyboar, line 1: This appears unused. Remove?

```suggestion
```

wesleyboar (approved):
Tested. Nothing breaks. Looks safe.

I don't immediately know how to use it, but it is required by https://github.com/TACC/tup-cms/tree/task/TUP-250--taccsite_custom-static-content which also works, and I do understand the use case of.

PR #522, opened 2022-07-26
fix(taccsite_ui): do not let django serve patterns

wesleyboar:
## Overview

Do **not** let Django serve UI patterns demo.

⚠️ Do not merge until we have another solution. Instead, cherry-pick from here before deploying other `task/fp-1499-cms-pattern-library--` branches (and revert when done deploying). (Yes, it's a mess.)

<details>

Letting Django server the demo files **might** have been the cause of a performance issue. And, they should be served with a dedicated static file server anyway.

<details>

## Related

- [FP-1499](https://jira.tacc.utexas.edu/browse/FP-1499)

## Changes

Remove Django-based static file hosting (not "staticfiles" but independent static assets).

## Testing

I deployed this on pprd.ecep.tacc.utexas.edu. The UI pattern library demo was unavailable (so any potential performance issue cannot occur).

## Screenshots

N/A

### DIFF
--- a/taccsite_cms/settings.py
+++ b/taccsite_cms/settings.py
@@ -198,12 +198,6 @@ TACC_BLOG_SHOW_CATEGORIES = True
 TACC_BLOG_SHOW_TAGS = True
 
 ########################
-# TACC: UI PATTERN DEMO
-########################
-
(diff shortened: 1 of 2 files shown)


wesleyboar:
A solution to deploy ui pattern demo will be revisited after new tup cms, which will give us the architecture to deploy it more sensibly.

PR #535, opened 2022-08-11
fix(settings): TUP-321 make LDAP_ENABLED work

wesleyboar:
## Overview

Make the `LDAP_ENABLED` setting actually do something.

<details>I did not find when it lost its effect (or if it ever had effect), but Jake and I suspect it does nothing. I do not see the setting used anywhere in Core-CMS code. But H.P. can login to tup-cms with her LDAP creds, even though we disabled that (for sure).</details>

## Related

None.

## Changes

- use `LDAP_ENABLED` to conditionally update and set settings

## Testing

(I'm not familiar enough with LDAP nor Django to test this locally. I am not focused enough to write better remote testing steps.)

1. Deploy on a server: https://dev.tup.tacc.utexas.edu/
2. Try to login with LDAP credentials.
3. Verify you cannot login.

## UI

N/A

### DIFF
--- a/taccsite_cms/_settings/auth.py
+++ b/taccsite_cms/_settings/auth.py
@@ -1,0 +1,41 @@
+"""Configure django authentication"""
+
+########################
+# DJANGO
+# https://docs.djangoproject.com/en/2.2/ref/settings/#auth
+########################
+
+# To disable LDAP:
+#
+# 1. In custom or local settings, duplicate the AUTHENTICATION_BACKENDS setting.
+# 2. Remove the '...LDAPBackend' entry.
+# 3. If 'INCLUDES_CORE_PORTAL = False', remove '...CorePortalAuthBackend' entry.
+#
+# RFE: Use INCLUDES_CORE_PORTAL to toggle '...CorePortalAuthBackend'.
+
+# Requires django-auth-ldap ≥ 2.0.0
+LDAP_ENABLED = True
+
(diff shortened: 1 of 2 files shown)


rstijerina (approved):
LGTM. If it matters, looks like the django-auth-ldap package is outdated at 2.2, 4.1 being the latest

wesleyboar:
I could not login with LDAP (after deleting my manually created account with the same password).

H.P. has two passwords that access her account (LDAP and manual, same username). She should only have one (manual). I've asked her to login with both to confirm (because I am unfamiliar enough with auth that I want to ensure the specific problem—hers—that caused my effort is solved).

wesleyboar:
S.T.'s concern is noted:
> [FP-1771](https://jira.tacc.utexas.edu/browse/FP-1771): CMS: Updated Outdated LDAP Auth

wesleyboar:
This is not working. I believe this, because:

1. I deleted H.P.'s account via CMS.
2. I confirmed via command line—in container and python shell—that H.P.'s has no account.
3. I created a Django CMS account for H.P. via the CMS.
4. H.P. can login with her LDAP account password.
5. After her login, her Django CMS account was updated/merged to have LDAP account First Name and Last Name.

wesleyboar:
The purpose of this branch has changed, since **both** d9d1684 **and** J.R. found a quick way to disable auth in tup-cms.

PR #542, opened 2022-08-16
Add ipython to django shell

rstijerina:
## Overview

Adds the ipython package to the environment, via `poetry add ipython`

By default, django will use ipython for the django shell if it is installed:
https://docs.djangoproject.com/en/2.2/ref/django-admin/#shell

## Testing

1. Build image: `make build`
2. Start it up: `make start`
3. Bring up django shell from within container:
  ```
docker exec -it core_cms bash
./manage.py shell
```
4. Confirm ipython is installed.

## UI

**Before (without ipython):**
Screen Shot 2022-08-16 at 1 51 32 PM

**After (with ipython):**
Screen Shot 2022-08-16 at 1 57 54 PM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ description = "DjangoCMS backend for the TACC ACI-WMA Core-CMS Codebase."
 authors = ["TACC-WMA <wma-portals@tacc.utexas.edu>"]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = ">=3.8,<4.0"
 aldryn-apphooks-config = "0.6.0"
 aldryn-boilerplates = "0.8.0"
 aldryn-common = "1.0.5"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pyproject.toml [hunk 1]
```diff
@@ -5,7 +5,7 @@ description = "DjangoCMS backend for the TACC ACI-WMA Core-CMS Codebase."
 authors = ["TACC-WMA <wma-portals@tacc.utexas.edu>"]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = ">=3.8,<4.0"
```
rstijerina, line 8: The Dockerfile is based on python-3.8; updated here to reflect that

rstijerina:
Side note: the pyproject.toml file has mostly exact version pins, rather than carrot pins to accept bugfix/patch-level version updates.

wesleyboar (approved):
Tested. Works with a custom docker-compose.

PR #531, opened 2022-08-04
chore: fp-1903 sync tup-ui login component changes to core-styles

wesleyboar:
## Overview

Updates based on styles moved to, and lessons learned from, tup-ui (via https://github.com/TACC/tup-ui/pull/47).

## Related

- [FP-1903](https://jira.tacc.utexas.edu/browse/FP-1903)
- requires https://github.com/TACC/Core-Styles/pull/41
- requires https://github.com/TACC/Core-CMS-Resources/pull/158

### Test Setup

1. `npm ci`
2. `npm run build --project=core-cms && docker exec -it core_cms python manage.py collectstatic --no-input`
    - where `core-cms` is your local CMS `container_name` ([alternatives](https://github.com/TACC/Core-CMS/wiki/How-to-Use-a-Custom-Docker-Compose-File))

Watch for, and report any errors, during testing.

### Changes, Steps, Pictures

<details><summary>ignore font files copied form Core-Styles</summary>

1. `git check-ignore taccsite_cms/static/site_cms/fonts/BentonSans-*` (from root of repo clone)
1. Verify BentonSans font files are ignored.
    ignore font files copied form Core-Styles

</details>
<details><summary>install Core-Styles at relevant branch</summary>

1. `npm list @tacc/core-styles`
2. Verify output has `git+ssh://git@github.com/TACC/Core-Styles.git#`.
      install Core-Styles at relevant branch

</details>
<details><summary>copy font files from Core-Styles</summary>

1. Verify four BentonSans font files are present in `taccsite_cms/static/site_cms/fonts`.
2. Verify BentonSans fonts are loaded.
    - You can do this via Developer Tools' Network tab.
        copy font files from Core-Styles

</details>
<details><summary>adjust path to fonts to be absolute</summary>

1. Verify BentonSans fonts are loaded from `/static/site_cms/fonts/`.
    - You can do this via Developer Tools' Network tab's resource panel's Preview tab.
        adjust path to fonts to be absolute

</details>
<details><summary>use variable for button font-size</summary>

1. Verify BentonSans fonts are loaded from `/static/site_cms/fonts/`.
    - You can do this via Developer Tools' Elements tab's style panel's Computed tab.
        use variable for button font-size

</details>
<details><summary>ensure certain form fields are only as wide as necessary</summary>

1. Create an instance of a "Form" plugin.
2. Add a Time field.
3. Add a Radio field.
4. Verify these fields are only as wide as necessary to fit their content (input value, input icon, label text).
[...]

PR #555, opened 2022-10-10
fix: FP-1891 add missing font

wesleyboar:
## To Do

- [x] Resolve conflicts
- ~~Test~~
- ~~Screenshot~~

## Overview / Changes

Load missing font that is added in TACC/Core-CMS-Resources#166.

## Related

- [FP-1891](https://jira.tacc.utexas.edu/browse/FP-1891)
- requires https://github.com/TACC/Core-CMS-Resources/pull/166

## Testing

1. Open a page with this branch.
2. Change a normal text paragraph to italic (`font-style: italic`).
3. Verify text becomes italic, but **not** bolder (`font-weight: 500`).

## UI

…

## Notes

~~The font is in Frontera resources (in `taccsite_custom`), because that is where the Benton Sans fonts are (because I was in a rush N years ago, just like I am today). When Frontera is migrated to [TACC/Core-CMS-Custom](https://github.com/TACC/Core-CMS-Custom), I will be forced to fix this (because font family will fail on **all** other Core-CMS sites, Frontera and not).~~

_**Update**: As of #531, fonts come from https://github.com/TACC/Core-Styles instead of Frontera._


taccsite_cms/static/site_cms/css/src/_imports/generics/fonts.css
```diff
@@ -29,6 +29,14 @@ Styleguide Generics.Fonts
   font-display: swap;
 }
```
wesleyboar, line 31: Self, please remove this line, to show the relationship of Regular to RegularItalic.

```suggestion
```

wesleyboar:
As of https://github.com/TACC/Core-CMS/pull/531, fonts come from [TACC/Core-Styles](https://github.com/TACC/Core-Styles) instead of Frontera. This PR is moot. Its Core-CMS-Resources is closed too.

PR #577, opened 2022-12-15
Task/update example cms defaults

taoteg:
## Overview

Creating an out-of-the-box ready-to-go example workflow for the Core-CMS in SAD mode.

## Related

- None.

## Changes

Updated the configs so the example-cms assets are all present and isolated from other projects.

## Testing

1. Create symlinks to the example files in the cloned repo (_./core-cms/)_: 

```
cd taccsite_cms 
ln -s settings_custom.example.py settings_custom.py
ln -s settings_local.example.py settings_local.py
ln -s secrets.py secrets.default.py
```

2. Run `make example`
3. Docker exec into the cms container and run the django `migrate`, `collectstatic` and `createsuperuser` tasks to complete the CMS setup.
4. Open the browser to `localhost:8000`, login as the admin user you created and use the portal.

## UI

Standard Core CMS UI.

## Notes

The settings_custom.py files included in the repos seem to be out of date, with both deprecated and missing values in them. The example workflow is intended to use the default example settings files exactly as they configured.


wesleyboar (commented):
I skimmed. I am excited to try this out. I like the changes I see.

_What I don't get yet (and maybe I will when I test and compare) is the difference between `docker-compose.example-cms.yml` and `docker-compose.dev.yml`._

taoteg:
I didn't want to disrupt anything that already existed, in case anyone has a workflow that depends on files that currently exist. The relationship between these files is:  

- `docker-compose.yml` --> `docker-compose.dev.yml`
- `docker-compose.example.yml` --> `docker-compose.example-cms.yml`

These `example` files alter the `extends` configuration (in the former) and the `image` configuration (in the latter), building the `example-cms:latest` image and starting it with default settings values as a SAD implementation.

wesleyboar:
I lost my unread GitHub notification for this PR. I am hoping this comment adds a new one.

wesleyboar (approved):
I tried this. It works.
Thank you for this clearer way to start.
I appreciate the backwards-compatibility.

Related—but not blocking—I don't think News should installed by default. Your thoughts?

PR #581, opened 2023-01-10
feat(dev/tup-cms): django3, blog/news styles + related patterns

wesleyboar:
## Overview

A dev branch for all new TUP CMS work.

<details>

This branch is used to build images for https://github.com/TACC/tup-ui/tree/main/apps/tup-cms. This lets me deploy everything so I can test "New Stuff A" without removing "New Stuff B". I will continuously deploy this branch for CMS changes. I keep this branch up to date merely by `git merge other/branch-name` as updates from other branches come.

</details>

## Related / Changes / Testing / UI

- [TUP-507](https://jira.tacc.utexas.edu/browse/TUP-507)
- requires https://github.com/TACC/Core-CMS-Resources/pull/176
- includes:
    - [v3.12.0-alpha.1: Django 3; Redesign Blog/News + Related Patterns](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.1)
    - [v3.12.0-alpha.2: Fix v3.10 Accent Color From Purple Back to Blue](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.2)
    - [v3.12.0-alpha.3: Install Image Gallery Plugin](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.3)
    - [v3.12.0-alpha.4: Gallery Updates, Style Fixes, taccsite_custom Update](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.4)
    - [v3.12.0-alpha.5: Remove demdata-cms, Merge v3.11.3](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.5)
    - [v3.12.0-alpha.6: Display a.img-fluid as Inline Block](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.6)
    - [v3.12.0-alpha.7: Core-Styles v2.9.1 to v2.11.0](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.7)
    - [v3.12.0-alpha.8: Re-Sync dev/tup-cms, main, taccsite_custom](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.8)
    - [v3.12.0-alpha.9: Remove Migrated Projects from Submodule](https://github.com/TACC/Core-CMS/releases/tag/v3.12.0-alpha.9)
    - #674

PR #619, opened 2023-03-31
feat(css): change CMS default accent color

wesleyboar:
## Status

This is unnecessary if https://github.com/TACC/Core-CMS/pull/581 is merged.

> **Info**
> Update: Unnecessary, because #660 was merged.

## Overview

Change CMS default accent color from purple to blue.

## Related

- mimics #617
    <sup>in an easy way that works for `main` branch without touching TACC/Core-Styles</sup>

## Changes

- **added** global accent color variables with value specific to CMS

## Testing

1. On Core CMS instance (or a CMS with minimal difference), verify link colors are blue not purple.

## UI

…

### DIFF
--- a/taccsite_cms/static/site_cms/css/src/_imports/settings/color.css
+++ b/taccsite_cms/static/site_cms/css/src/_imports/settings/color.css
@@ -2,6 +2,16 @@
 
 :root {
 
+  /* Accent */
+  --global-color-accent--x-light: #caddfe;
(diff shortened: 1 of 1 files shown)


wesleyboar:
This will be unnecessary if https://github.com/TACC/Core-CMS/pull/581 is merged.

wesleyboar:
Unnecessary, because of #660.

PR #623, opened 2023-04-10
feat(templates): extract assets_delayed_custom js ❌

wesleyboar:
## Overview

~~Extract `assets_delayed_custom` template into smaller templates, so when TACC/tup-ui (unfortunately) overwrites the file, it can easily recreate its original content.~~

**This changes `assets_core_delayed`, which is unnecessary.**

## Related

- required by ___

## Changes

- **changes** `assets_delayed_custom.html` to load markup from separate templates

## Testing

Noop.

0. Load page and inspect element.
1. Verify `assets_delayed_custom.html` original content is still loaded.

## UI

core-cms pr 623


wesleyboar:
This pull request is solving a problem that does not exist. https://github.com/TACC/tup-ui/pull/181 messes with `assets_`**`custom`**`_delayed.html`, not `assets_`**`core`**`_delayed.html`.

PR #579, opened 2023-01-02
feat(server): fp-1952 tighter ES network settings

wesleyboar:
## Overview

Set up Elasticsearch network to limit exposure of the software on the network.

## Related

- [FP-1952](https://jira.tacc.utexas.edu/browse/FP-1952)
- mimics https://github.com/TACC/Core-Portal/pull/752

## Changes

- limit network to only localhost

## Testing

1. Run development environment using these new settings.
2. Test that Elasticsearch still works:
    1. Add a CMS page **and** confirm re-index occurs and is accurate.[^1]
    2. Publish a CMS page **and** confirm re-index occurs and is accurate.[^1]
    4. Delete a CMS page **and** confirm re-index occurs and is accurate.[^1]
    5. Add text to a CMS page and verify page appears in a search.

[^1]: You can check container logs for a message saying that re-index has occurred. You can check http://localhost:9201/cms-dev-cms/_search to confirm search index has expected page content.

## UI

Skipped.

## Notes

<details><summary>Archived</summary>

Elasticsearch [**Networking** reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#modules-network) suggests [**`network.host`**](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#common-network-settings) should be [`_local_`](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#network-interface-values) instead of `0.0.0.0`.

**Update**: But doing so would interfere with Docker cross-container communication.
</details>

PR #627, opened 2023-05-01
bugfix: security for example-cms docker compose

wesleyboar:
## Overview

Port a security fix to the example docker compose files.

## Related

- mimics #579

## Changes

- **add** host IP for port set in docker compose

## Testing & UI

Skipped. Should be a noop change. Please correct me if I am mistaken.


taoteg (approved):
Nice.

PR #630, opened 2023-05-02
fix: TUP-477, set X_FRAME_OPTIONS like django 2

wesleyboar:
## Overview

Fix problem caused by [Django 3 protection against clickjacking](https://docs.djangoproject.com/en/3.0/ref/clickjacking/).

## Related

- [TUP-477]

[TUP-477]: https://jira.tacc.utexas.edu/browse/TUP-477

## Changes

- **added** `X_FRAME_OPTIONS = 'SAMEORIGIN'`

## Testing

Test [TUP-477].

## UI

Skipped.

### Help Wanted

1. Is there an option that only sets `SAMEORIGIN` where we know we need it, and otherwise defaults to `DENY`?
2. Is that option more secure? And sensible?


wesleyboar:
Works on [today's deploy of this](https://tacc-team.slack.com/archives/C02GF2X2AS3/p1683127951084949).

PR #631, opened 2023-05-03
fix: TUP-477, set X_FRAME_OPTIONS like django 2 (for TUP CMS)

wesleyboar:
Mimics #630, but for #626 instead of #625.


wesleyboar:
Works on [today's deploy of this](https://tacc-team.slack.com/archives/C02GF2X2AS3/p1683127951084949).

PR #665, opened 2023-06-22
docs(README): improvements

wesleyboar:
## Overview

Improve the README.

## Related

None

## Changes

- **changed** README

## Testing

1. Read through the changes.
2. Evaluate the improvements.

## UI

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ These projects store project-specific resources in the `taccsite_custom` submodu
 7. [Collect static files][django-static] for Django:
 
     ```bash
-    python manage.py collectstatic
+    python manage.py collectstatic --no-input
     ```
 
     _[If `DEBUG` is set to `True`, then this is automated via `python manage.py runserver`.][django-static-serve-dev]_
@@ -164,7 +164,7 @@ These projects store project-specific resources in the `taccsite_custom` submodu
 
     The CMS admin site should now be accessible at http://localhost:8000/admin .*
 
-    _You may log in as the superuser created via the `createsuperuser` command in an earlier step._
+    _You should log in as the superuser created via the `createsuperuser` command in an earlier step._
 
     <sub>\* Or at the port defined in a `docker-compose.custom.yml`.</sub>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,416 +1,178 @@
...
 
-<sub>\* You should run these commands in the container __from `/code/`__. _See [Running Commands in Container](#running-commands-in-container)._</sub>
-
-### Changing Custom Resources
-
-If you need to get the latest files into `/taccsite_custom` from [Core CMS Resources]:
-
-1. [How to Change Submodule Branch Commit](https://github.com/TACC/Core-CMS/wiki/How-to-Change-Submodule-Branch-Commit)
+3. Django CMS:
+    1. Open http://0.0.0.0:8000/.
```
fnets, line 93: I'm stepping a little out of my depth here, but should we use `localhost:8000`? Wouldn't the IP depend on local configuration?

fnets (approved):
I don't know much about the non-APCD CMS sites, but this makes sense

README.md
wesleyboar:
```suggestion
    1. Open http://localhost:8000/.
```

Yes. This was copied. It was `localhost`. Thanks.

I'll update the repos I copied from too.

wesleyboar: My"Yes" is to whether we should use `localhost:8000`. I **believe** the IP depends on the local configuration, which is I **think** [is 127.0.0.1 thus localhost](https://github.com/TACC/Core-CMS/blob/main/docker-compose.dev.yml#L8-L10).

wesleyboar: Wes, you updated the sources:
- https://github.com/TACC/Core-CMS-Resources/blob/docs/readme-improvements/README.md#start-project
- https://github.com/TACC/Core-CMS-Custom#start-project

wesleyboar:
I had been mimicking https://github.com/TACC/Core-Portal README format as closely as possible, but CMS is ready for its own README format, and this documentation update coincides well with https://github.com/TACC/Core-CMS/releases/tag/v3.12.0 **and** the deprecation of https://github.com/TACC/Core-CMS-Resources [^1] **and** the increasing migration to https://github.com/TACC/Core-CMS-Custom [^1].

[^1]: These repos have similar style README's to this PR's README.

PR #672, opened 2023-07-14
task WI-46: Setup CMS using django-csp

chandra-tacc:
## Overview
To become fully secure site, adding detailed Content Security Policy is needed.
This uses django-csp to enable csp.

## Related

* [WP-46](https://jira.tacc.utexas.edu/browse/WP-46)
* similar to https://github.com/TACC/Core-Portal/pull/829
* delayed/replaced by https://github.com/TACC/Camino/pull/32

## Changes
This PR adds CSP headers for
* font src
* script src
* style src
* connect src
Also, ensure the current script tags use nonce. 
The setup right now is in "report only" mode to allow for opt-in and fully functional app.

## Testing

1. Validated the site using UI and reducing console warnings.

## UI
No UI change.

## Notes:
At this point, due to possibly breaking the app due to CSP, this PR is in draft mode. Other mitigations are deployed via https://github.com/TACC/Camino/pull/32


wesleyboar:
Seems important, but I have not heard a request to revisit this, so I'm marking this "medium" priority.

wesleyboar (commented):
Do `<link>` elements actually use `nonce`?

- [W3.org states](https://www.w3.org/TR/CSP2/#script-src-the-nonce-attribute) "Nonce sources require a new nonce attribute to be added to both [script](https://www.w3.org/TR/html5/scripting-1.html#the-script-element) and [style](https://www.w3.org/TR/html5/document-metadata.html#the-style-element) elements."
- The [MDN `nonce`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) page does not mention `<link>`. Only `<script>` and `<style>`.

I've created a merge conflict resolution for review — #745 — but it has a bug.

PR #693, opened 2023-08-21
enhance: improve admin UI permissions error text

wesleyboar:
## Overview

Improve confusing text error message in the Django CMS Admin UI.

## Related

None.

<details><summary>Backstory</summary>

When trying to give staff member Publish permission on [ECEP](https://ecepalliance.org/), I hit confusing error message.

<sup>I got angry at Django CMS. I thought about proposing better error text on Django CMS proper. Then I realized I don't know if I should edit their `msgstr` in their `en` po file, or just change in my own app. I remembered how much all of Django CMS text sucks, but also how unfamiliar I am with what most of it means, and did not trust myself to improve DJango CMS at the source. Especially, because I am not running—thus not experiencing—their latest software; so, I am out-of-date.<sup>

</details>

## Changes

- **changed** four error messages.

## Testing

1. Open http://0.0.0.0:8000/admin/cms/globalpagepermission/add/.
2. For any Group or User, give/check only the permission "Can publish".

## UI

| before | after |
| - | - |
| user perm text - before | user perm text - after |


taoteg (approved):
LGTM! Clearer action for users.

PR #695, opened 2023-08-22
hotfix: set SESSION_COOKIE_SECURE = True

rstijerina:
## Overview

Sets `SESSION_COOKIE_SECURE = True` to resolve a flag set by an ISO security audit.

## Testing

1. In the chrome devtools, go to Application -> Cookies -> and ensure that the "Secure" column for the "sessionid" cookie is checked. See screenshot below

## UI
Screenshot 2023-08-22 at 9 37 19 AM

### DIFF
--- a/taccsite_cms/settings.py
+++ b/taccsite_cms/settings.py
@@ -41,6 +41,9 @@ CEP_AUTH_VERIFICATION_ENDPOINT = 'localhost'  # 'https://0.0.0.0:8000'
 # https://docs.djangoproject.com/en/3.0/ref/clickjacking/#how-to-use-it
 X_FRAME_OPTIONS = 'SAMEORIGIN'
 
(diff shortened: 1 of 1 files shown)


taoteg (approved):
LGTM.

wesleyboar (approved):
I am unavailable to test yet. I approve it because I don't anticipate an issue. And Django says it should be `True` (even though default it to `False`).

https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-SESSION_COOKIE_SECURE

wesleyboar:
If every CMS needs this, then this should be done off on a branch off of v3.11.5 tag, so I can make a v3.11.6 release. Most sites are running v3.11. Core-CMS does **not** have a _stable_ v3.12 yet.

PR #717, opened 2023-09-27
docs: manage scripts and styles

wesleyboar:
## Overview

How I _currently_ manage scripts and styles, **but** not necessarily how I _should_.

## Related

- any future work by unfamiliar developers

## Changes

- **added** document about adding scripts and styles

## Testing

N/A

## UI

Read https://github.com/TACC/Core-CMS/blob/docs/styles-and-scripts/docs/styles-and-scripts.md.

### DIFF
--- a/docs/develop-project.md
+++ b/docs/develop-project.md
@@ -4,11 +4,16 @@ Read [Django CMS User Guide] for CMS user instructions.
 
 ## Table of Contents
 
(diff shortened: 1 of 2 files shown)


wesleyboar:
I merge this in for the upcoming Core-CMS 4.0 release, for which I am also improving more documentation.

We can revisit how the scripts and styles management **should** be next. At least no, I document how it **is**.

PR #727, opened 2023-10-05
redirect admin login view to use the base portal login

jarosenb:
## Overview
Replaces the admin login view with a redirect to the base `/login`. This prevents users from seeing a login screen that isn't wired up to Tapis/LDAP.

## Testing

1. go to `<portal URL>/admin` while unauthenticated
2. Enter your credentials and check that you're redirected to the admin screen.


wesleyboar (changes requested):
I requested this, but didn't think of the repercussions and edge cases.

1. Must be conditional — probably via `INCLUDES_CORE_PORTAL` — for S.A.D[^1] CMS's.
2. CMS admins must be prepared to take care of users who had a CMS-only account.[^2]
3. CMS admins could be locked out, because they might be admins only via a CMS account.

[^1]: Stand-alone CMS (no portal).
[^2]: For some CMS's, that may be the CMS admins themselves.

wesleyboar:
> **Important**
> We **do** want this, just not until:
> - SAD CMS login is still possible.
> - CMS admins use TACC accounts.
> - CMS admins are ready to support users.

PR #741, opened 2023-11-02
Update dockerfile to speed up builds

jarosenb:
## Overview
Update the Dockerfile to remove Node deprecation warnings in CI and streamline Python dependency installation

## Related
- [WI-63](https://jira.tacc.utexas.edu/browse/WI-63)

## Changes
- Use a separate Node build step to handle style building
- Update Poetry config to use the container's system Python
- expose the Poetry binary when Core-CMS is used as a base image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,5 @@
 FROM python:3.8-buster as python-base
-
 LABEL maintainer="TACC-ACI-WMA <wma_prtl@tacc.utexas.edu>"
-
-ARG DEBIAN_FRONTEND=noninteractive
-
-ENV PYTHONUNBUFFERED 1
-
-# https://python-poetry.org/docs/configuration/#using-environment-variables
-ENV POETRY_VERSION=1.4.0 \
-    POETRY_HOME="/opt/poetry" \
-    POETRY_VIRTUALENVS_IN_PROJECT=true \
-    POETRY_NO_INTERACTION=1 \
-    PYSETUP_PATH="/opt/pysetup" \
-    VENV_PATH="/opt/pysetup/.venv"
-
-# prepend poetry and venv to path
-ENV PATH="$POETRY_HOME/bin:$VENV_PATH/bin:$PATH"
-
-FROM python-base as builder-base
 
 RUN apt-get update && apt-get install -y \
     build-essential python3-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,46 +9,46 @@ RUN apt-get update && apt-get install -y \
 
 RUN pip3 install --upgrade pip setuptools wheel
 
+ARG DEBIAN_FRONTEND=noninteractive
```
rstijerina, line 12: Do we want to keep this arg at the top? I think it comes into play with the `apt-get install` step

jarosenb: Oh good catch, this doesn't seem to impact the build at all but I've re-ordered the steps to match what was being run originally.

Dockerfile
```diff
@@ -1,72 +1,50 @@
...
 # Install Poetry - respects $POETRY_VERSION & $POETRY_HOME
 RUN curl -sSL https://install.python-poetry.org | python3 -
-
+RUN mkdir /code
 # copy project requirement files here to ensure they will be cached.
-WORKDIR $PYSETUP_PATH
-COPY pyproject.toml poetry.lock ./
-
+COPY pyproject.toml poetry.lock /code/
+WORKDIR /code
```
wesleyboar, line 24: This reduces steps from 3 to 2, with the same effect, correct?

Dockerfile
```diff
@@ -1,72 +1,50 @@
...
-# install node
-RUN curl -sL https://deb.nodesource.com/setup_17.x | bash -
-RUN apt-get install -y nodejs
+FROM node:18 as node_build
+COPY package.json package-lock.json /code/
+WORKDIR /code
+RUN npm ci
 
-COPY --from=builder-base $PYSETUP_PATH $PYSETUP_PATH
+COPY . /code/
```
wesleyboar, line 39: Ah, I think I get it. First copy node dep's list in and build, then copy the rest in. This way, changes to the rest of the code don't trigger layer change thus don't cause `npm ci` for unrelated code changes.

Dockerfile
```diff
@@ -1,72 +1,50 @@
...
-RUN mkdir /code
-COPY . /code
-WORKDIR /code
-
-# build assets
-RUN npm ci
-ARG PROJECT_NAME
-ARG BUILD_ID
-RUN npm run build --project=$PROJECT_NAME --build-id=$BUILD_ID
+COPY --from=node_build /code/ /code
```
wesleyboar, line 50: And this copies the result of `npm ci` and `npm run build …` to the final layer's `/code`.

- (new) `/node_modules`
- (unchanged) `/package.json` and `/package-lock.json`
- (populated) `/css/build` and `/taccsite_ui/dist`

wesleyboar (commented):
Looks good.

(As a separate task) Would it be inappropriate to run `python manage.py collectstatic` via the Dockerfile? _I think this would save ~~devs and~~[^1] CI that step in client apps._

[^1]: Already unnecessary for devs — if they have `DEBUG` set to `True` — because [`urls.py#L82-L83`](https://github.com/TACC/Core-CMS/blob/v4.1.0/taccsite_cms/urls.py#L82-L83) (since 1c326b9).

---

Also, I should test this PR thoroughly.

- [x] Core-CMS local
- [ ] Core-CMS remote
- [ ] Core-CMS-Resources local
- [ ] Core-CMS-Resources remote
- [ ] Core-CMS-Custom local
- [ ] Core-CMS-Custom remote
- [ ] tup-ui local
- [ ] tup-ui remote

Dockerfile
```diff
@@ -1,72 +1,50 @@
...
-
-ENV PYTHONUNBUFFERED 1
-
-# https://python-poetry.org/docs/configuration/#using-environment-variables
-ENV POETRY_VERSION=1.4.0 \
-    POETRY_HOME="/opt/poetry" \
-    POETRY_VIRTUALENVS_IN_PROJECT=true \
-    POETRY_NO_INTERACTION=1 \
-    PYSETUP_PATH="/opt/pysetup" \
-    VENV_PATH="/opt/pysetup/.venv"
```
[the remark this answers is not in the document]
taoteg: If I am grokking correctly, because line 11 (in the changed column -->) is now using the container's default Python, there is no need to define the `PYSETUP_PATH`. The `POETRY_HOME` and `VENV_PATH` are no longer being used in the `PATH` definition either because the default locations would now be correct. So all three values can get dropped.

taoteg (approved):
LGTM! Nice update.

wesleyboar (approved):
I think I understand this well now. Based on my understanding, [I tested enough](https://github.com/TACC/Core-CMS/pull/741#pullrequestreview-1727747306).

wesleyboar:
I cleaned up Dockerfile comments in f932d5f.

PR #775, opened 2023-12-14
task/tup-557 Fixes drop cap articles for Core-CMS

R-Tomas-Gonzalez:
## Overview

Fixes drop cap articles for Core-CMS

## Related

- [TUP-557](https://tacc-main.atlassian.net/browse/TUP-557)

## Changes

Quick fix, changes css to initial-letter. NOTE. This will not work on Firefox at the time of developing.

## Testing

1. Check http://localhost:8000/static/ui/components/detail/s-drop-cap.html on different browsers. Currently works for chromium and safari.

## UI

| safari |
| - |
| Screenshot 2023-12-14 at 4 47 05 PM |

| chrome |
| - | 
| Screenshot 2023-12-14 at 4 47 27 PM |


taccsite_cms/static/site_cms/css/src/_imports/tools/x-drop-cap.css
```diff
@@ -1,19 +1,12 @@
 .x-drop-cap {
-  /* Drop the letter */
-  float: left;
   /* Estimate horizontal spacing to match magazine design */
   /* WARN: Firefox calculates `margin-…` differently here, so use `padding-…` */
   padding-right: 0.125em; /* approx: `1em` of normal text */
 
   font-size: 6em; /* approx: 3 lines, of normal text, tall */
```
wesleyboar, line 6: **Opportunity for less code.** Having font-size seems unnecessary. I tried without it and `initial-letter` resized the font for me. I did need to change the padding-right, though, because `em` became based off of parent's font-size.

taccsite_cms/static/site_cms/css/src/_imports/tools/x-drop-cap.css
```diff
@@ -1,19 +1,12 @@
...
-  /* CAVEAT: Also has effect in Safari and Chrome/Edge, but not enough */
-  padding-top: 0.0625em;
-  /* (Safari v13, Chrome/Edge v80) Nudge letter up more until it meets top of normal text */
-  /* MAGIC: Firefox does not need this */
-  /* SEE: https://bugzilla.mozilla.org/show_bug.cgi?id=1233271 */
-  /* MAGIC: Has no effect on Firefox */
-  /* SEE: https://bugzilla.mozilla.org/show_bug.cgi?id=290125 */
-  line-height: 0.65;
+  /* Not yet available on Firefox */
+  initial-letter: 3;
```
wesleyboar, line 9: **Uh oh.** This renders poorly on Firefox. Please use [Progressive Enhancement](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) to do this this on browsers that support it, but not on browsers that don't. CSS offers [Feature Detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection) via @supports.

taccsite_cms/static/site_cms/css/src/_imports/tools/x-drop-cap.css
```diff
@@ -1,19 +1,12 @@
...
+  /* Support for Safari */
+  -webkit-initial-letter: 3;
```
wesleyboar, line 11: **Hm.** Safari uses a serif font. Not awful... Have you found any solution or bug report?

R-Tomas-Gonzalez: Possible known issue: https://caniuse.com/css-initial-letter
Screenshot 2023-12-15 at 3 45 30 PM

wesleyboar: K. Safari 9 is too old. No concern.

> Safari implementation is incomplete. Does not allow applying web fonts to the initial letter.

And that explains why the CMS's web font is not used. I haven't thought of a workaround. This bug can stay. If M.S. vomits, we'll just drop `-webkit-initial-letter`.

taccsite_cms/static/site_cms/css/src/_imports/trumps/s-drop-cap.css
```diff
@@ -1,5 +1,15 @@
-@import url('../tools/x-drop-cap.css');
+@supports (initial-letter: 3){
+  .s-drop-cap > p:first-of-type::first-letter {
+    padding-right: 1em; /* approx: `1em` of normal text */
```
wesleyboar, line 3: This seems **exactly** `1em` of normal text now. 😀

```suggestion
    padding-right: 1em;
```

taccsite_cms/static/site_cms/css/src/_imports/trumps/s-drop-cap.css
```diff
@@ -1,5 +1,15 @@
...
+    initial-letter: 3;
+  }
+}
 
-.s-drop-cap > p:first-of-type::first-letter {
-  @extend .x-drop-cap;
+@supports (-webkit-initial-letter: 3) {
+  .s-drop-cap > p:first-of-type::first-letter {
+    padding-right: 1em; /* approx: `1em` of normal text */
```
wesleyboar, line 10: This seems **exactly** `1em` of normal text now. 😀

```suggestion
    padding-right: 1em;
```

wesleyboar (commented):
Thanks for this. I'll test and approve today or Monday.

wesleyboar (commented):
Noticed one thing while testing.

taccsite_cms/static/site_cms/css/src/_imports/trumps/s-drop-cap.css
```diff
@@ -1,5 +1,15 @@
...
+    margin-left: -1em; /* remove apprx. space on left side of text */
```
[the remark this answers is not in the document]
R-Tomas-Gonzalez: Ahh gotcha. I'll remove this. I was concerned with the tiny bit of padding you see on the drop-cap there without the margin. But if we're okay with it, there's no reason to put it in.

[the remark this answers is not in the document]
wesleyboar: Sensible. I notice the tiny bit of padding. M.S. does very much like to align things…

But my theory is that the extra space is built into the letter of that font, and any such extra space may be different per letter and per font (especially given Safari's incomplete implementation), thus attempts to avoid the space will not always work. Feel free to test the theory, offer alternatives, or ask M.S. what he'd prefer.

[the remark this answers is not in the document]
wesleyboar: I wanted to make sure I wasn't making stuff up. I tested my theory on just Arial font. Different letters need different love.

| D | Z | A | P |
| - | - | - | - |
| D | Z | A | P |

wesleyboar (approved):
Code approved.

PR #778, opened 2023-12-21
chore: tup-677 css/django/vite interoperability

wesleyboar:
## Overview

Edit CSS so that Django and Vite do not use it unexpectedly.

> [!WARNING]
> This change causes errors collecting them as static files, so all `collectstatic` commands have had `--ignore` param added to avoid the error. This is intended to be a **temporary** workaround.

## Related

- [TUP-677](https://tacc-main.atlassian.net/browse/TUP-677)

<details><summary>Pull Requests</summary>

- https://github.com/TACC/tup-ui/pull/400 requires:
    - https://github.com/TACC/Core-Styles/pull/285
    - https://github.com/TACC/Core-CMS/pull/778 requires:
        - https://github.com/TACC/Core-CMS-Resources/pull/197
        - https://github.com/TACC/Core-Styles/pull/285

</details>

### CSS
- **refactored** comments that suggest to `@import` code
- **refactored** absolute `/static` paths to be relative
- **changed** core-styles version

### Static Files / Cache
- **added** `STORAGES` setting to "fix a cache problem"
- **changed** static file config to ignore only `*src/*.css"` not all of `src`
- **documented** additional parameter for `collectstatic` command

## Testing

0. **Uncertain.** See https://github.com/TACC/tup-ui/pull/398.
1. Verify in a WYSIWYG editor that an instance of "Styles" → "Block Styles" → "Island" is styled.

## UI

c-island

### How to FIx The Issue w/ Font

> In some of these font files `url(fonts/...)` needs to be `url(../fonts/..)` e.g. https://github.com/TACC/Core-Styles/blob/v2.22.4/dist/settings/font.css.
> — (paraphrased) @jarosenb

> The paths to fonts have — in the past — been manipulated via https://github.com/TACC/Core-Styles/blob/v2.22.4/src/.postcssrc.base.yml#L37-L43.
> — (paraphrased) @wesleyboar

### DIFF
--- a/taccsite_cms/django/contrib/staticfiles_custom/apps.py
+++ b/taccsite_cms/django/contrib/staticfiles_custom/apps.py
@@ -12,5 +12,5 @@ class TaccStaticFilesConfig(StaticFilesConfig):
         'CVS', '.*', '*~',
         # Added by TACC
         # WARNING: Ignores these from non-TACC static's also
-        'src', 'README.md'
+        'README.md', "*src/*.css"
     ]
--- a/taccsite_cms/settings.py
+++ b/taccsite_cms/settings.py
@@ -323,6 +323,15 @@ ui_demo_dir = os.path.join(BASE_DIR, 'taccsite_ui', 'dist')
 if os.path.exists(ui_demo_dir):
     STATICFILES_DIRS += (('ui', ui_demo_dir),)
 
+STORAGES = {
+    "default": {
+        "BACKEND": "django.core.files.storage.FileSystemStorage",
+    },
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
+    },
+}
+
 # User Uploaded Files Location.
 MEDIA_URL = '/media/'
 MEDIA_ROOT = os.path.join(DATA_DIR, 'media')
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "taccsite_custom"]
 	path = taccsite_custom
 	url = git@github.com:TACC/Core-CMS-Resources.git
-	branch = task/37-per-site-assets
+	branch = chore/tup-677-css-django-vite-interoperability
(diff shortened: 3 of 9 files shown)


taccsite_cms/django/contrib/staticfiles_custom/apps.py
```diff
@@ -12,5 +12,5 @@ class TaccStaticFilesConfig(StaticFilesConfig):
         'CVS', '.*', '*~',
         # Added by TACC
         # WARNING: Ignores these from non-TACC static's also
-        'src', 'README.md'
+        'README.md', "*src/*.css"
```
wesleyboar, line 15: **Explanation:**
> by excluding all "src" folders it breaks certain CMS stylesheets that reach into their own src/ folders for icons/fonts. With this change we now only exclude CSS files in src/ folders, which is the more correct behavior.
> — @jarosenb

taccsite_cms/settings.py
```diff
@@ -323,6 +323,15 @@ def gettext(s): return s
...
     STATICFILES_DIRS += (('ui', ui_demo_dir),)
 
+STORAGES = {
+    "default": {
+        "BACKEND": "django.core.files.storage.FileSystemStorage",
+    },
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
+    },
+}
```
wesleyboar, line 333: **Introduction:**
We now deviate from [default settings](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES) in order to use `ManifestStaticFilesStorage`.

> The purpose of this storage is to keep serving the old files in case some pages still refer to those files, e.g. because they are cached by you or a 3rd party proxy server.
> — [Docker 4.2 @ `ManifestStaticFilesStorage`](https://docs.djangoproject.com/en/4.2/ref/contrib/staticfiles/#manifeststaticfilesstorage)

taccsite_custom
wesleyboar, line 1: **See** https://github.com/TACC/Core-CMS-Resources/pull/197.

wesleyboar (commented):
> [!WARNING]
> This change causes errors collecting them as static files, so all `collectstatic` commands have had `--ignore` param added to avoid the error. This is intended to be a **temporary** workaround.

PR #783, opened 2024-01-09
feat: TV3-181 allow remote LOGO and FAVICON

wesleyboar:
## Overview

- Allow adding remote URLs for logo and favicon.
- Support `TACC_LOGO` and `TACC_FAVICON`.
- Deprecate `LOGO` and `FAVICON`.

## Related

- [TV3-181](https://tacc-main.atlassian.net/browse/TV3-181)
- required by https://github.com/TACC/Core-CMS-Custom/pull/257
- expected by https://github.com/TACC/Core-CMS-Custom/pull/260

## Changes

- **added** `is_remote` boolean for FAVICON settings
- **added** simple protocol parse logic for LOGO filename
- **added** non-static asset load for LOGO and FAVICON settings

## Testing

1. Load Core-CMS **sans** `taccsite_cms/settings_custom.py`.
2. Verify logo is default — white "PORTAL".
3. Verify favicon is default — white star on blue field.
4. Load Core-CMS **with** `taccsite_cms/settings_custom.py`.
    <sub>`cp taccsite_cms/settings_custom.example.py taccsite_cms/settings_custom.py`</sub>
5. Restart server.
6. Verify logo is PTDataX's — "DATAX".
7. Verify favicon is PTDataX's — multi-colored "X".
8. Edit `taccsite_cms/settings.py`, uncomment `LOGO` setting.
9. Edit `taccsite_cms/settings.py`, uncomment `FAVICON` setting.
10. Restart server (if necessary).
11. Verify logo is default — white "PORTAL".
12. Verify favicon is default — white star on blue field.
13. Re-test https://github.com/TACC/Core-CMS-Custom/pull/260.

## UI

| Standard | Custom | Deprecated |
| - | - | - |
| 2 3 Standard | 6 7 Custom | 11 12 Deprecated |

## Notes

Used on client via https://github.com/TACC/Core-CMS-Custom/pull/260.


taccsite_cms/templates/header_logo.html
```diff
@@ -5,7 +5,7 @@
   {% with filename=logo|index:1 selectors=logo|index:2 targeturl=logo|index:3 targettype=logo|index:4 accessibility=logo|index:5 corstype=logo|index:6 visibility=logo|index:7 %}
     {% if visibility == "True" %}
     <a id="header-logo" class="navbar-brand {{className}}" href="{{ targeturl }}" target="{{ targettype }}">
-      <img class="portal-logo {{ selectors }}" src="{% static filename %}" crossorigin="{{ corstype }}" alt="{{ accessibility }}" >
+      <img class="portal-logo {{ selectors }}" src="{% if "://" in filename %}{{ filename }}{% else %}{% static filename %}{% endif %}" crossorigin="{{ corstype }}" alt="{{ accessibility }}" />
```
rstijerina, line 8: Would the new `is_remote` param be possible to key off here for static vs non-static?

wesleyboar: **Yes, with effort.** The FAVICON is a dict, so adding `is_remote` was easy. The LOGO is a list, so adding `is_remote` could be cumbersome.

### Ideas

**A.** Add _optional_ entry to LOGO setting. Add _careful_ logic to avoid "index out of range" error in template. Test _thoroughly_.
**B.** Add new TACC_LOGO dict setting. Deprecate LOGO list setting. Have template accept both, until all sites are updated.
**C.** Add LOGO_IS_REMOTE setting. Maintain both.

I prefer **B**.

rstijerina: Though cumbersome now to refactor, I agree B sounds best for longevity

wesleyboar: I committed changes to support B:
- `TACC_LOGO` and `TACC_FAVICON` are standard[^1]
- `LOGO` and `FAVICON` take precedence, for backwards-compatibility

[^1]: Other. newer Core-CMS–specific settings are prefixed with `TACC_`, to mirror `DJANGO_` and `CMS_` and `BLOG_`.

I tested, but haven't taken screenshots yet.

wesleyboar: Okay. UI for testing on local Core-CMS. Re-deployed pprd PTDataX.

rstijerina: You may have addressed this elsewhere, but curious on your thoughts of using `PORTAL_` as a prefix instead of `TACC_`

wesleyboar: My thoughts are _okay_… I've noticed the problem with `TACC_`, but `TACC_` is consistent... I will try to write a Python function to cleanly support `TACC_` and `PORTAL_` until `TACC_` is deprecated. Or, I give up and just do it for these. I want to give more attention on this. I will return.

rstijerina: I'm totally fine with merge now, change later in this case if more convenient.

taccsite_cms/templates/assets_custom.html
```diff
@@ -12,8 +12,8 @@
 
 
 <!-- Custom Site Assets: Favicon. -->
-{% with settings.FAVICON as favicon %}
-<link rel="icon" href="{% static favicon.img_file_src %}" type="image/x-icon" />
+{% with settings.TACC_FAVICON as favicon %}
+<link rel="icon" href="{% if favicon.is_remote %}{{ favicon.img_file_src }}{% else %}{% static favicon.img_file_src %}{% endif %}" type="image/x-icon" />
```
rstijerina, line 16: Would you be able to make this change for Core-Portal as well?

rstijerina (approved):

taccsite_cms/templates/header_logo.html
wesleyboar: I put some thought into it.

1. **We CAN support new prefix LATER**,
    - via some slim Python code I don't know how to write yet.
2. **I think prefix should not be `PORTAL_`**, because:
    - clarity of Core-Portal vs Core-CMS settings seems important
    - it propagates ill-defined word "Portal"[^1]
3. **The prefix should NOT be `CMS_`**, because:
    - Django CMS uses that prefix
4. **I am tempted to use `CORE_CMS_`**, because:
    - the `CMS_` shows Core-CMS and Django CMS relationship
    - the `CORE_` shows Core-CMS and Core-Portal relationship


[^1]: TACC's website-portal-projects are, each as a whole, called "portals" even though they are a combination of a CMS **website** — the "(usually) not behind the login" aspect — and a Portal **app** —the "(mostly) behind the login" aspect.

wesleyboar:
I will return with a PR that supports prefixes `TACC_` and `CORE_CMS_` (and/or something else we decide), so that the confusing `TACC_LOGO` and `TACC_FAVICON` do not propagate.

taccsite_cms/templates/header_logo.html
wesleyboar: [CMD-79: Prefix CMS Settings w/ `CORE_CMS_` not `TACC_`](https://tacc-main.atlassian.net/browse/CMD-79)

And, **ugh**, I will not use slim Python code for dynamic variables to support both prefixes. I will use a manual list of aliases, for legibility. Chat GPT made a valid point when I asked it for help... AI is not useless in programming (just limited).

wesleyboar:
See [CMD-79: Prefix CMS Settings w/ `CORE_CMS_` not `TACC_`](https://tacc-main.atlassian.net/browse/CMD-79) next.

taccsite_cms/templates/assets_custom.html
rstijerina: 1 and 2: There is the risk of file collision if both portal and cms use the same static directory. Likelihood does seem low. Is there an advantage to both using the same static url?
3: Can you expand on this? I'm not sure what you're referring to exactly

taccsite_cms/templates/header_logo.html
rstijerina: I would say we are a bit beyond point 2, as in we are invested in that prefix/terminology. I'm not against migrating references of PORTAL to CORE_PORTAL across the board though, but that would be a separate task in my view.

wesleyboar: Yeah, I spent last night and this morning trying not to submit to a `PORTAL_`. **I submit!**[^1] I've renamed the [CMD-79](https://tacc-main.atlassian.net/browse/CMD-79) ticket, and will use `PORTAL_`.

[^1]: From one perspective, it is all a portal. The CMS is a small piece. From a different perspective, CMS surrounds the webpages and the portal. From yet another perspective the CMS and Portal are both optional pieces of a whole… that is still called Portal.

taccsite_cms/templates/assets_custom.html
wesleyboar: **1. & 2.** I can't think of one.
**3.** Unrelated topic I was forcing into the conversation through logic I did not expose. Please never mind for this topic.

**Yes, I can do it.**

On Portal, would the value `/static/...` (to load from Core-CMS) use `is_remote = TRUE`, because it's not using Core-Portal's `{% static %}`?

rstijerina: > On Portal, would the value /static/... (to load from Core-CMS) use is_remote = TRUE, because it's not using Core-Portal's {% static %}?

Ah I see, the portal already kinda handles this in that it doesn't default to using `% static %`. Think it's still needed?

taccsite_cms/templates/header_logo.html
wesleyboar: - #792

taccsite_cms/templates/assets_custom.html
wesleyboar: _I forgot about this. Back now. And reading it without the flu fogging my brain._

I think Portal does not need `{% static %}` option for favicon. It works for remote and CMS assets with its code as is. _To update Portal would be for consistency. A welcome change, but a rabbit hole. The divergence occurred and is great now._

<details><summary>Details (for those who didn't read the thread)</summary>

Adding `{% static %}` and `is_remote` option to Portal adds complexity. The Portal already assumes favicon is not a "local" asset, thus assumes it will be given a non-Portal-static-asset URL. If existing hardcoded `/static/` values concern us, we can add something like `CMS_STATIC_DIR` in default settings.

https://github.com/TACC/Core-Portal/blob/v3.5.0/server/portal/templates/base.html#L11-L13

```py
    {% if settings.PORTAL_ICON_FILENAME %}
    <link rel="icon" href="{{ settings.PORTAL_ICON_FILENAME }}">
    {% else %}
```

</details>

wesleyboar: @taoteg will test this theory on DigitalRocks. 😀

PR #792, opened 2024-01-31
refactor: cmd-79 rename tacc cms settings

wesleyboar:
## Overview

Reword `TACC_` settings as `PORTAL_` settings.

<sup>Because (A) Core-CMS is used by clients, so "TACC_LOGO" is a weird setting (because they have their own logo) and (B) The CMS and Portal are aspects of a website that is called Portal in TACC-speak (and I am done [grumbling against it](https://github.com/TACC/Core-CMS/pull/783#discussion_r1471742829)).</sup>

## Status

- [x] test
    - [ ] https://github.com/TACC/tup-ui/pull/416
    - [x] https://github.com/TACC/Core-CMS-Custom/pull/265
    - [x] https://github.com/TACC/Core-Portal-Deployments/pull/43

## Related

- [CMD-79](https://tacc-main.atlassian.net/browse/CMD-79)
- continues #795
- required by:
    - https://github.com/TACC/tup-ui/pull/416
    - https://github.com/TACC/Core-CMS-Custom/pull/265
    - https://github.com/TACC/Core-Portal-Deployments/pull/43
- includes:
    - #837

## Changes

- **renamed** settings

## Testing & UI

https://github.com/TACC/Core-Portal-Deployments/pull/43


taccsite_cms/_settings/auth.py
```diff
@@ -14,9 +14,9 @@
 # 1. Set 'LDAP_ENABLED = False'.
 # 2. Duplicate the 'AUTHENTICATION_BACKENDS' setting.
 # 3. Remove its '...LDAPBackend' entry.
-# 4. If 'INCLUDES_CORE_PORTAL = False', remove '...CorePortalAuthBackend' entry.
+# 4. If 'PORTAL_IS_CORE = False', remove '...CorePortalAuthBackend' entry.
```
taoteg, line 17: The term `PORTAL_IS_CORE` is ambiguous to me. While some portals have the full `CORE-PORTAL`, some only have the `CORE-CMS`, but ALL portals are `CORE`.

Wouldn't `USES_PORTAL` be clearer as to the relationship being defined here?

taccsite_cms/settings.py
```diff
@@ -288,9 +287,14 @@ def gettext(s): return s
 # TACC: PORTAL
 ########################
 
-INCLUDES_CORE_PORTAL = True
-INCLUDES_PORTAL_NAV = True
-INCLUDES_SEARCH_BAR = True
+PORTAL_IS_CORE = True
```
taoteg, line 290: See comment above on `AUTH_USES_CORE` idea.

wesleyboar: Responded. I will remember from the prior discussion to update this instance as we decide.

taccsite_cms/_settings/auth.py
wesleyboar: I do not disagree. I offer facts so we can find the best name.

1. Not all Portals are Core-Portal. TUP uses Core-CMS, but not Core-Portal.
2. In this context, `AUTH_USES_PORTAL` is clearer.
3. There is another context where `PORTAL_IS_CORE` is used: [`/taccsite_cms/urls.py`](https://github.com/TACC/Core-CMS/blob/feat/cmd-79-support-new-core-cms-settings-prefix/taccsite_cms/urls.py#L31-L41).

Perhaps this should become two settings:
- `PORTAL_MANAGES_AUTH` (or `AUTH_USES_PORTAL`, but I am trying to prefix all settings with `PORTAL_`[^1])
- `PORTAL_IS_CORE`

Thoughts?

[^1]: Because that seems to be what other CMS apps do e.g. `djangocms-blog` uses `BLOG_`.
[^2]: So we know at a glance the origin of the setting is a TACC Portal.

wesleyboar (commented):
Replied.

taoteg:
`PORTAL_MANAGES_AUTH` works for me, it has the `PORTAL_` prefix and explicitly states what it does.
`PORTAL_IS_CORE` feels too ambiguous imo. _Question: What property/aspect of a portal configuration is this supposed to be indicating if not that the AUTH source is the django container?_

Aside: I do so which we had called `Core-Portal` something less overloaded (`Core-Workbench`, `Core-Dashboard`, `Core-Django`, `Core-Workspace`, `literally anything else that would not conflate the use of portal`), but it is what it is now...

wesleyboar:
I changed `PORTAL_IS_CORE` to `PORTAL_IS_TACC_CORE_PORTAL`.

<details>

Goal is clarity and accuracy. I see the setting is **not** _solely_ used to describe Core-Portal managing auth. It also describes code that Core-Portal expects but is not necessary for all portals (e.g. TUP, or a future one with a better way to share header) nor related to auth.

</details>

Perhaps we revisit `PORTAL_MANAGES_AUTH` as a setting if we make CMS support LDAP.

wesleyboar:
I tested this on **every** Core-CMS website —

- https://github.com/TACC/Core-Portal-Deployments/pull/43

— and the only bug was because of upgrading ECEP to v4.9+.

> [!IMPORTANT]
> Merging based on approval of https://github.com/TACC/Core-Portal-Deployments/pull/43.

PR #793, opened 2024-01-31
task/cmd-78 Allow designers to toggle off drop-caps to articles

R-Tomas-Gonzalez:
## Overview

Django Admin _would_ be the ideal place to put this feature. 
However, Django admin is specifically used as a FORM BASE system, so this would be the wrong place to try and put this feature.

Allowing the designer to toggle off the article switch (IMO) would have to be done from the Developer side with a setting to make it easier... In my limited amount of time, I could not figure out a way to add a button to the Admin section, without creating _what Django calls_ a new application.

There was a point where I added a new button to the header bar, but again, I do not believe this is the right place to add this feature.

## Related
- [CMD-78](https://tacc-main.atlassian.net/browse/CMD-78)


R-Tomas-Gonzalez:
Admin section is the ideal place for this button, but not the ideal place for Django. Womp. Closing PR. Leaving this here for documentation.

wesleyboar:
Excellent. This will prevent another from falling down the same rabbit hole.

PR #794, opened 2024-02-01
deps/ipython-8.21.0: Update ipython for Python 3.11 compatibility

jarosenb:
## Overview
- update iPython to prevent errors when accessing the shell
- remove Pillow dependency that was causing issues with local install. We don't use Pillow directly so we should let Poetry solve for its version based on our other deps.


pyproject.toml
```diff
@@ -80,7 +80,6 @@ lxml = "4.9.4"
 mccabe = "0.6.1"
 packaging = "20.4"
 parso = "0.7.1"
-Pillow = "9.0.1"
```
wesleyboar, line 83: **Concern Resolved.** For posterity, Pillow is **not** being **un**installed. It still remains, upgraded in fact, in the `.lock` file, to v10. Pillow is a direct requirement of [easy-thumbnails](https://github.com/SmileyChris/easy-thumbnails/blob/2.8.5/setup.py#L36), so I understand that Pillow need not be listed in our `pyproject.toml`.

wesleyboar (approved):
Cool.

PR #815, opened 2024-03-12
CMD-73 Remove Taccsite Blockquote Plugin from Text Editor

R-Tomas-Gonzalez:
## Overview

Removes blockquote plugin from text editor

## Related

[CMD-73](https://tacc-main.atlassian.net/browse/CMD-73)

## Changes

Changes `taccsite_cms/contrib/taccsite_blockquote/cms_plugins.py` -> `text-enabled = True` to `text-enabled = False`

## Testing

1. Create a text plugin
2. In the plugin, there's a cms plugin dropdown
3. Make sure there is no Blockquote plugin under TACC Site

## UI

Screenshot 2024-03-12 at 5 54 08 PM

### DIFF
--- a/taccsite_cms/contrib/taccsite_blockquote/cms_plugins.py
+++ b/taccsite_cms/contrib/taccsite_blockquote/cms_plugins.py
@@ -19,7 +19,7 @@ class TaccsiteBlockquotePlugin(CMSPluginBase):
     render_template = 'blockquote.html'
 
     cache = True
-    text_enabled = True
+    text_enabled = False
     allow_children = False
 
     fieldsets = [


wesleyboar (commented):
What would happen to a page if a Taccsite Blockquote plugin instance were already added **within/via** a Text plugin instance, then the CMS is updated to have this change, and the page is loaded? _I hope nothing, but I wouldn't be surprised if the CMS vomits a complaint.[^1]_

[^1]: I fear Django CMS code may say "Wait. `text_enabled = False` and there is one inside the Text? No… I just can't with this user. 500. I'm out."

R-Tomas-Gonzalez:
> What would happen to a page if a Taccsite Blockquote plugin instance were already added **within/via** a Text plugin instance, then the CMS was updated to have this change, and the page was loaded? _I hope nothing, but I wouldn't be surprised if the CMS vomits a complaint.[1](#user-content-fn-1-15d90dec7a8af05e1c58c1ac729f05c5)_
> 
> ## Footnotes
> 1. I fear Django CMS code may say "Wait. `text_enabled = False` and there is one inside the Text? No… I just can't with this user. 500. I'm out." [↩](#user-content-fnref-1-15d90dec7a8af05e1c58c1ac729f05c5)

Tested backwards compatibility here. It doesn't scream. But although it doesn't scream, I think it's something we keep in mind.

Screen Recording: https://tacc-team.slack.com/archives/D060KCUQVL6/p1710333901862999?thread_ts=1710284123.284459&cid=D060KCUQVL6

wesleyboar (approved):

PR #861, opened 2024-08-07
feat: secrets.example.py

wesleyboar:
## Overview

Create a `secrets.example.py`.

## Related

- similar to https://github.com/TACC/Core-CMS/pull/577
- similar to [Core-Portal's `settings_secret.example.py`](https://github.com/TACC/Core-Portal/blob/v3.7.0/server/portal/settings/settings_secret.example.py)

## Changes

- **added** `secrets.example.py`
- **documented** creation of `secrets.py`
- **documented** creation of `settings_custom.py`

## Testing & UI

N/A

### DIFF
--- a/taccsite_cms/secrets.example.py
+++ b/taccsite_cms/secrets.example.py
@@ -4,16 +4,6 @@
 
 SECRET_KEY = 'CHANGE_ME'
 
-# Specify allowed hosts or use an asterisk to allow any host.
-# ALLOWED_HOSTS = ['hostname.tacc.utexas.edu', 'client.org'] # Dev/Prod/Etc
-ALLOWED_HOSTS = ['0.0.0.0', '127.0.0.1', 'localhost', '*']   # Local
-
-########################
-# TACC: PORTAL
-########################
-
-CEP_AUTH_VERIFICATION_ENDPOINT = 'http://django:6000'
-
 ########################
 # STORAGE
 ########################
--- a/taccsite_cms/secrets.example.py
+++ b/taccsite_cms/secrets.example.py
@@ -1,0 +1,48 @@
+########################
+# DJANGO
+########################
+
+SECRET_KEY = 'CHANGE_ME'
+
+# Specify allowed hosts or use an asterisk to allow any host.
+# ALLOWED_HOSTS = ['hostname.tacc.utexas.edu', 'client.org'] # Dev/Prod/Etc
+ALLOWED_HOSTS = ['0.0.0.0', '127.0.0.1', 'localhost', '*']   # Local
+
+########################
+# TACC: PORTAL
+########################
+
+CEP_AUTH_VERIFICATION_ENDPOINT = 'http://django:6000'
+
+########################
+# STORAGE
+########################
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'PORT': '5432',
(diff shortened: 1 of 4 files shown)


taoteg (changes requested):
A number of these settings are placed in the `*.settings_local.py` file rather than `secrets.py`. 

Specifically:
- `ALLOWED_HOSTS`
- `CEP_AUTH_VERIFICATION_ENDPOINT`

Reason: These settings do not contain sensitive credentials.

Current Example: https://github.com/TACC/Core-Portal-Deployments/blob/main/ami/camino/prod.cms.settings_local.py

Can we split these into the appropriate files (using this same setup but including the `settings_local.py` file as well) so the usage matches the deployment environment?

wesleyboar:
Thanks, yes! Will do. I am glad to have your eyes on this.

taccsite_cms/secrets.example.py
```diff
@@ -0,0 +1,38 @@
...
+########################
+# SEARCH
+########################
+
+ES_AUTH = 'username:password'
+ES_HOSTS = 'http://elasticsearch:9200'
+ES_INDEX_PREFIX = 'cms-dev-{}'
+ES_DOMAIN = 'http://localhost:8000'
+
+HAYSTACK_CONNECTIONS = {
```
wesleyboar, line 31: @taoteg, should **all** of these stay in `secrets.py`?

- The `ES_AUTH` value seems secret.
- The `ES_…` values are expected by `HAYSTACK_CONNECTIONS`.

wesleyboar (commented):
@taoteg I moved secret values to `settings_local.py`. Also, please see my question about `ES_` values.

taccsite_cms/secrets.example.py
taoteg: Yes, keep in secrets. While only the `ES_AUTH` value is truly sensitive, it is easier to keep them all together for tracking purposes (and avoid leaking any seemingly harmless info about internal portal configs). 

We might should have another review as to "what values live in which files" again, it has been a while AFAIK since we evaluated the specific location of values distributed across the defaults/custom/local/secrets files.

wesleyboar: Moving `ES_…` values would break ES if `HAYSTACK…` is in any of those files.

I've hit loading order complexity before. I solved it via [new section in `settings.py` **after** settings/secrets import](https://github.com/TACC/Core-CMS/blob/v4.15.2/taccsite_cms/settings.py#L808-L814).

I think the complexity of `ES_` in many files is "not worth it". I feel safe to do so, but I defer to you.[^1]

[^1]: Would you feel compelled to update such files on servers or in Core-Portal-Deployments?

taoteg: I 100% agree that it is "not worth it" due to the second-order effects.

> Would you feel compelled to update such files on servers or in Core-Portal-Deployments?

I don't think any `ES_` values live anywhere _except_ in the `secrets.py` files, so you have to update the file on the individual host (and add any changes to the Stache entry, naturally).

wesleyboar: K. So, good to go this PR?

taoteg (approved):
LGTM!

taccsite_cms/secrets.example.py
taoteg: my bad, yes - approved!

PR #953, opened 2025-05-21
added nonce to analytics

taoteg:
## Overview

Checking for delta between `main` and dev branch.

## Related

…

## Changes

Added nonce for analytics requests.

## Testing

1. Build and deploy Core-CMS in prod.cep to verify that analytics are still being tracked properly. 
2. Will have to either:
2A. Find the logging from ITS scans being run against the host to see if the flagged issue has been resolved in the security scans.
2B. Find a way to validate the nonce directly (some kind of bash check I suspect).

## UI

None.

### DIFF
--- a/taccsite_cms/templates/analytics.html
+++ b/taccsite_cms/templates/analytics.html
@@ -5,7 +5,7 @@
     </script>
   {% else %}
     <!-- Global site tag (gtag.js) & Google Analytics Code for Site.  -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id={{ settings.GOOGLE_ANALYTICS_PROPERTY_ID }}"></script>
+    <script async nonce="CSP_NONCE" src="https://www.googletagmanager.com/gtag/js?id={{ settings.GOOGLE_ANALYTICS_PROPERTY_ID }}"></script>
 
     <!-- Analytics Code -->
     <script>


happycodemonkey (approved):
looks good to me

taccsite_cms/templates/analytics.html
```diff
@@ -5,7 +5,7 @@
     </script>
   {% else %}
     <!-- Global site tag (gtag.js) & Google Analytics Code for Site.  -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id={{ settings.GOOGLE_ANALYTICS_PROPERTY_ID }}"></script>
+    <script async nonce="CSP_NONCE" src="https://www.googletagmanager.com/gtag/js?id={{ settings.GOOGLE_ANALYTICS_PROPERTY_ID }}"></script>
```
wesleyboar, line 8: This is a static string value. I think [a nonce is suppose to be a random value per HTTP response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#nonces).

1. Is this a test? If so, why merge instead of build image from branch?
2. Why do we now need a nonce? Is it because Google Analytics changed something, and we should  ["avoid using the CSP `unsafe-inline` directive"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce#description) as a solution?

    > […] If you don't need nonce, don't use it. […] Always try to take full advantage of [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) protections and avoid nonces or unsafe inline scripts whenever possible.
    > — [MDN: `nonce`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce#description)

> [!NOTE]
> @chandra-tacc's attempt at this was https://github.com/TACC/Core-CMS/pull/672. It probably made a dynamic value, but added `nonce` on more elements than seemed necessary.

wesleyboar (commented):
I have concerns.

PR #978, opened 2025-08-12
chore: safely unregister picture plugin

wesleyboar:
## Overview

Do not assume unregistering the original picture plugin will succeed. 

## Related

- improves https://github.com/TACC/Core-CMS/releases/tag/v4.36.0

## Changes

- **adds** `try`/`catch` around plugin unregister

## Testing / UI

Skipped.

## Notes

A.I. wanted to do this originally. I took it out. Maybe it is safer to not assume. Unsure, in this case. Core-CMS does indeed use original picture plugin, and will continue to do so.

### DIFF
--- a/taccsite_cms/djangocms_picture/extend.py
+++ b/taccsite_cms/djangocms_picture/extend.py
@@ -1,3 +1,7 @@
+import logging
+
+logger = logging.getLogger(f"portal.{__name__}")
+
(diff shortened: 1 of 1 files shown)


wesleyboar:
I still don't like it. I want to know, before release, if plugin unregister will fail, because websites might depend on plugin being unregistered and then new one registered.

PR #912, opened 2025-02-24
Quick: Use non-slim jQuery dep to fix djangocms-forms incompatibility

jarosenb:
## Overview
The jQuery form plugin used by djangocms-forms is **not** compatible with jQuery slim (see: https://github.com/jquery-form/form). This diff replaces jquery-slim with the full-fat version, which includes the required ajax config.

## Related

- fixes bug in https://github.com/TACC/tup-ui/pull/499 [^1]
- caused by #901 [^1]

[^1]: Note by @wesleyboar

PR #1051, opened 2026-01-06
Quick: Use non-slim jQuery dep to fix djangocms-forms incompatibility

fnets:
## Overview
The jQuery form plugin used by djangocms-forms is not compatible with jQuery slim (see: https://github.com/jquery-form/form). This diff replaces jquery-slim with the full-fat version, which includes the required ajax config.
…

## Related
Re-enables fix from previous PR: https://github.com/TACC/Core-CMS/pull/912

## UI

| before | after |
| - | - |
| dev-cep BEFORE | dev-cep AFTER |


fnets:
> #1053 branch was merged into this PR's branch, to reduce diff. This PR would still be merged into `stable`. I re-open this PR.

Thanks, Wes!

PR #1095, opened 2026-02-17
Update settings.py to set CSRF_COOKIE_SECURE

jarosenb:
## Overview
Set CSRF_COOKIE_SECURE to true. This was getting flagged on automated security scans.


wesleyboar (approved):
I **added a [commit](https://github.com/TACC/Core-CMS/pull/1095/changes/93922d0)** to continue to support local development.

I **tested local instance login**[^1] on Firefox, Chrome, and Safari.

[^1]: Using all value combinations.

PR #1107, opened 2026-02-26
enhance: restore SRC_ROOT to setup-cms.sh

wesleyboar:
## Overview

Restore `SRC_ROOT` to `setup-cms.sh`. TACC/Core-CMS-Custom changes this variable. Having it here introduces parity.

## Related

- reverts non-functional change from #1106


wesleyboar:
Qodo, they are the same. If they differ, build output would be in… well, for this repo, maybe still project root. Okay, I yield. Changed in c9250398.

PR #1116, opened 2026-03-11
Quick: specify unique csrftoken name for cms

rstijerina:
## Overview

CMS and Portal containers were using identical csrftoken cookie names, causing weird issues. This changes the name of the cms cookie, which is the simplest change to make them distinct.

## Testing

1. Confirm csrftoken is now named "cmscsrftoken" in browser


wesleyboar (approved):
Login works. CMS cookie has new name.

Image

PR #1211, opened 2026-07-20
refactor(breadcrumbs): move script-dependent `<style>` to `<script>`

wesleyboar:
## Overview

The "disabled" second-level breadcrumb link's style override lived in a standalone `<style>` block, separated from the script that strips its `href` and creates the need for that override — making it look like generic breadcrumb styling worth moving to a shared design system, when it's actually specific to this script's DOM change.

## Related

- discussion: https://github.com/TACC/Core-Portal/pull/1332#discussion_r3606778994

## Changes

- **moved** script-dependent `<style>` to `<script>`

## Testing

1. `make start`
2. Visit a page nested at least two levels deep in the CMS nav.
3. Verify the second breadcrumb link still renders normally (full opacity, inherited color) instead of looking disabled.
4. Verify it's still non-clickable when its target isn't a direct nav link, same as before.

## UI

Screenshot 2026-07-20 at 18 17 36

## Notes

No dependency or cross-repo changes needed — an earlier plan to migrate this CSS into Core-Styles (and Core-Portal) turned out to be based on a false premise: the style only compensates for CMS-specific JS that Core-Portal's breadcrumbs component doesn't have.

### DIFF
--- a/taccsite_cms/templates/nav_cms_breadcrumbs.html
+++ b/taccsite_cms/templates/nav_cms_breadcrumbs.html
@@ -1,15 +1,6 @@
 {# @var className #}
 {% load menu_tags %}
 
-{# To adjust style of "disabled" link #}
-{# NOTE: Not using `block`, so markup is together for devs in live markup #}
-<style>
-  .s-breadcrumbs a:not([href]) {
-    opacity: 1;
-    color: unset;
-  }
-</style>
-
 <nav class="s-breadcrumbs  {{className}}" id="cms-breadcrumbs">
 
   {# To support structured data, item* attributes are used #}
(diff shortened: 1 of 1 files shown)


taccsite_cms/templates/nav_cms_breadcrumbs.html
```diff
@@ -40,6 +31,13 @@
     const isDirectNavLink = href && cmsNav && cmsNav.querySelector(
       'a.nav-link:not(.dropdown-toggle)[href="' + href + '"]'
     );
-    if (!isDirectNavLink) secondLink.removeAttribute('href');
+    if (!isDirectNavLink) {
+      secondLink.removeAttribute('href');
+
+      {# To undo Core-Styles' disabled-link look; link is "current page" #}
```
wesleyboar:
```suggestion
      {# To undo Core-Styles' disabled look for parent root pages #}
```

